PR #7, opened 2025-07-08
added possibility to name the user keys

pasquale95:


pasquale95:
@liran-funaro i added a function to give names to the user keys with `cryptogen`. This function is needed for a better keys distribution and it was left as a TODO in `cryptogen`, but already implemented within the `fabric-smart-client` (see https://github.com/hyperledger-labs/fabric-smart-client/blob/45f4d482bcbfcb5a4175c6e55c0f49246cfc5e1e/integration/nwo/cmd/cryptogen/cryptogen.go#L401-L416).
I added this function here so that we can use `cryptogen` directly instead of `fsccli cryptogen`.

pasquale95:
Hi @liran-funaro, thanks for the suggestions. I have applied them.
The additional functionality aims to give `cryptogen` to capability to create named certificates for users within a peer organization as it is currently done with the `peers`.
Before this fix, it was impossible to specify the name of a user, leading to the production of user crypto material that looked like:
```text
users/
  Admin@org1.example.com/
  User1@org1.example.com/
```
With this additional functionality, if the user specifies a user with a Name:
```yaml
    Users:
      Count: 1
	  Specs:
	  	- Name: testuser
```
it will generate:
```text
users/
  Admin@org1.example.com/
  User1@org1.example.com/
  testuser@org1.example.com/
```
The possibility to give names to the user credentials is crucial to successfully handle the current deployment when using `cryptogen`, since the Ansible scripts use the hostname to identify the path where to locate the crypto material.

PR #12, opened 2025-09-08
export mock identity and identity deserializer

cendhu:
#### Type of change

- Improvement (improvement to code)

#### Description

Exporting mock objects to reuse code in fabric-x-committer for testing.

### DIFF
--- a/common/cauthdsl/cauthdsl.go
+++ b/common/cauthdsl/cauthdsl.go
@@ -24,7 +24,7 @@ var cauthdslLogger = util.MustGetLogger("cauthdsl")
 // passing them to this function for evaluation
 func compile(policy *cb.SignaturePolicy, identities []*mb.MSPPrincipal) (func([]msp.Identity, []bool) bool, error) {
 	if policy == nil {
-		return nil, fmt.Errorf("Empty policy element")
+		return nil, fmt.Errorf("empty policy element")
 	}
 
 	switch t := policy.Type.(type) {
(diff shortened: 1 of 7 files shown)


sampleconfig/configtx.yaml
```diff
@@ -69,9 +70,48 @@ Organizations:
...
                 Type: Signature
                 Rule: "OR('SampleOrg.member')"
+        OrdererEndpoints:
+            - "127.0.0.1:7050"
+
+    - &Org1
+        Name: Org1
+        SkipAsForeign: false
+        ID: Org1
+        MSPDir: crypto/Org1/peers/peer0.Org1/msp
```
liran-funaro: Why is it using the peer's MSP and not the Org's MSP?
The same for Org2.

liran-funaro (commented):
A few minor comments

liran-funaro (approved):
Why are some files `.crt`, `.key`, or just `priv_key`? And others use the `.pem`?
Since we use the PEM format in all of them, I suggest naming them all with `.pem`.

cendhu:
> Why are some files `.crt`, `.key`, or just `priv_key`? And others use the `.pem`? Since we use the PEM format in all of them, I suggest naming them all with `.pem`.

Yes. I did notice but the whole crypto folder was created by the cryptogen tool and I didn't rename or change the extension. Probably, the old cryptogen might have added `.pem` extension while the newer one does not.

liran-funaro (approved):

PR #39, opened 2025-12-10
[cryptogen] Adjust config block parameters for the loadgen

liran-funaro:
#### Type of change

- Breaking change

#### Description

In the method `CreateDefaultConfigBlockWithCrypto()`
- Receive a meta-namespace verification key directly instead of generating it.
- Receive the arma data directly instead of via a file.

#### Related issues

- required for https://github.com/hyperledger/fabric-x-committer/issues/188


cendhu (approved):
LGTM.

PR #53, opened 2026-01-09
replace msp.SerializedIdentity with protomsp.Identity

### LINKED ISSUE #43: replace all msp.SerializedIdentity with protomsp.Identity
The code currently inconsistently uses both msp.SerializedIdentity (which holds an MSP ID and certificate bytes) and protomsp.Identity (which holds an MSP ID and a Creator with either certificate bytes or a certificate ID).

We need to replace all occurrences of msp.SerializedIdentity with protomsp.Identity. This will allow us to update the IdentityDeserializer to convert only protomsp.Identity to msp.Identity. Additionally, we can update signedData to hold the protomsp.Identity struct directly, rather than storing the serialized bytes of msp.SerializedIdentity. This would help to avoid unnecessary serialization and deserialization of identity.

### LINKED ISSUE #44: Avoid unnecessary serialization and deserialization with protouitls.SignedData

cendhu:
#### Type of change

#### Description

This commit replaces all occurrences of msp.SerializedIdentity with the applicationpb.Identity except for the idemix.

Further, the identity in SignedData is changed to applicationpb.Identity rather than SerializedIdentity bytes to avoid unnecessary serialization/deserialization.

#### Related issues

 - resolves #43 
 - resolves #44


tools/pkg/identity/identity.go
```diff
@@ -19,7 +19,7 @@ type Signer interface {
 //
 // Serialize converts an identity to bytes.  It returns an error on failure.
 type Serializer interface {
```
tock-ibm: I prefer to have this interface be the same as before, i.e.
```
type Serializer interface {
	Serialize() ([]byte, error)
}
```

and have a second one with a different name, e.g.
```
type ExtendedSerializer interface {
	Serialize() ([]byte, error)
    SerializeWithIDOfCert() ([]byte, error)
}
```

then this also stays the same
```
type SignerSerializer interface {
   Signer
   Serializer
}
```

whereas an extended version has it all:
```
type ExtendedSignerSerializer interface {
   Signer
   ExtendedSerializer
}
```

This will make life a bit easier in the orderer and in all legacy code we bring from fabric.

What do you think?

cendhu: Anyway, I have made the requested changes to not block this PR for longer than necessary.

api/applicationpb/identity.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package applicationpb
```
liran-funaro: major: With this current design, the Orderer will have to use the `applicationpb` package to get the identity.
This breaks the Orderer's indifference to the application.
It makes more sense to move the identity to `protomsp` (should be `msppb`).
@tock-ibm What do you think?

common/cauthdsl/cauthdsl_test.go
```diff
@@ -14,31 +14,41 @@ import (
...
 
 func TestSimpleSignature(t *testing.T) {
 	t.Parallel()
-	policy := policydsl.Envelope(policydsl.SignedBy(0), signers)
+	policy := policydsl.Envelope(policydsl.SignedBy(0), signersBytes)
 
 	spe, err := compile(policy.Rule, policy.Identities)
 	if err != nil {
 		t.Fatalf("Could not create a new SignaturePolicyEvaluator using the given policy, crypto-helper: %s", err)
 	}
```
liran-funaro: minor: Can we adjust this to the proper form?
```suggestion
require.NoError(t, err, "Could not create a new SignaturePolicyEvaluator using the given policy, crypto-helper")
}
```

common/ledger/blkstorage/blockindex_test.go
```diff
@@ -369,7 +369,7 @@ func TestExportUniqueTxIDsErrorCases(t *testing.T) {
 	os.RemoveAll(testSnapshotDir)
 
 	// error during metadata file creation
-	fmt.Printf("testSnapshotDir=%s", testSnapshotDir)
+	fmt.Printf("testSnapshotDir=%s\n", testSnapshotDir)
```
liran-funaro: nit
```suggestion
	t.Logf("testSnapshotDir=%s", testSnapshotDir)
```

api/applicationpb/identity.go
tock-ibm: I agree!
Moreover - I wonder whether it is good practice to put general Go code in a package dedicated to protobuffs?

tock-ibm (approved):
LGTM

PR #64, opened 2026-01-27
 return error when no certificates exist in directory

### LINKED ISSUE #63: Return error when certificates are missing
In https://github.com/hyperledger/fabric-x-common/blob/b88e555fa9c77bbc5f9ccbbc1332705bcf13240c/msp/configbuilder.go#L210:
```
	cacerts, err := getPemMaterialFromDir(cacertDir)
	if err != nil || len(cacerts) == 0 {
		return nil, errors.WithMessagef(err, "could not load a valid ca certificate from directory %s", cacertDir)
	}
```

When certificates are missing, `getPemMaterialFromDir` returns a nil error and an empty list of certificates. 
As a result, calling `errors.Wrapf `or `errors.WithMessagef` with a nil error produces a nil error, and no error is propagated when calling `msp.GetLocalMspConfig(....)`. We need to propagate an error. 

This was fixed in the Fabric repo https://github.com/hyperledger/fabric/pull/5375, the same should be done here.

DorKatzelnick:
#### Type of change
Improvement to code

#### Description

The goal of this code change is to correctly return an error when the directory contains no certificates.

The function getPemMaterialFromDir returns a nil error and an empty list of certificates when the directory contains no certificates. As a result, calling errors.Wrapf or errors.WithMessagef with a nil error produces a nil error, and no error is propagated. The proposed fix addresses this issue by explicitly handling this case and ensuring that an appropriate error is returned.

#### Related issues

 - resolves #63

### DIFF
--- a/msp/configbuilder.go
+++ b/msp/configbuilder.go
@@ -180,8 +180,10 @@ func GetLocalMspConfig(dir string, bccspConfig *factory.FactoryOpts, ID string) 
 	}
 
 	signcert, err := getPemMaterialFromDir(signcertDir)
-	if err != nil || len(signcert) == 0 {
-		return nil, errors.Wrapf(err, "could not load a valid signer certificate from directory %s", signcertDir)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not load signing certificate from directory %s", signcertDir)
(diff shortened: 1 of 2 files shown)


liran-funaro (approved):
LGTM

PR #74, opened 2026-02-11
remove MetaNamespaceVerificationKey

### LINKED ISSUE #58: use LifecycleEndorsementPolicy instead of MetaNamespaceKey
The current configuration block relies on MetaNamespaceKey to define endorsement rules for creating namespaces. This needs to be replaced with LifecycleEndorsementPolicy to correctly align with the lifecycle management architecture in Fabric.

cendhu:
#### Type of change

#### Description

As we are planning to use LifecycleEndorsementPolicy instead of MetaNamespaceVerificationKey, we can remove it from the config block.

#### Related issues
 
  - resolves #58

### DIFF
--- a/common/channelconfig/application.go
+++ b/common/channelconfig/application.go
@@ -8,7 +8,6 @@ package channelconfig
 
 import (
 	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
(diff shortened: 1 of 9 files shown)


liran-funaro (approved):
LGTM

PR #98, opened 2026-04-07
remove golangci-lint from go.mod to avoid GPL dependency flagging

cendhu:
#### Type of change

- Dependency update

#### Description

Move golangci-lint installation from go.mod tool directive to scripts/install-dev-dependencies.sh to keep GPL-licensed transitive dependencies out of the module graph.


cendhu:
@liran-funaro I am merging this PR due to some urgent needs.

PR #103, opened 2026-04-19
Test block signature verifier with real policy

liran-funaro:
#### Type of change

- Test update

#### Description

- Test block signature verifier with real policy
- Move `testcrypto` package from `fabric-x-committer`

### DIFF
--- a/protoutil/blockutils_test.go
+++ b/protoutil/blockutils_test.go
@@ -17,9 +17,14 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/hyperledger/fabric-x-common/api/msppb"
+	"github.com/hyperledger/fabric-x-common/api/types"
+	"github.com/hyperledger/fabric-x-common/common/channelconfig"
 	configtxtest "github.com/hyperledger/fabric-x-common/common/configtx/test"
+	"github.com/hyperledger/fabric-x-common/common/policies"
+	"github.com/hyperledger/fabric-x-common/msp"
 	"github.com/hyperledger/fabric-x-common/protoutil"
 	"github.com/hyperledger/fabric-x-common/protoutil/mocks"
+	"github.com/hyperledger/fabric-x-common/utils/testcrypto"
 )
 
 var testChannelID = "myuniquetestchainid"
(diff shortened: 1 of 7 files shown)


protoutil/blockutils_test.go
```diff
@@ -508,6 +512,82 @@ func TestBlockSignatureVerifierByIdentifier(t *testing.T) {
...
+		Port:     7050,
+		MspId:    "SampleOrg",
+		Identity: certificateBytes,
+	}}
+
+	// Create a real cauthdsl policy that requires 1 signature from SampleOrg
+	policyEnvelope := policydsl.SignedByMspMember("SampleOrg")
+	policyProvider := cauthdsl.EnvelopeBasedPolicyProvider{Deserializer: testMSP}
+	policy, err := policyProvider.NewPolicy(policyEnvelope)
+	require.NoError(t, err)
```
tock-ibm: We need to test that we indeed trigger the BFT policy that requires 2f+1 out of 3f+1.
What's tested here is the Raft policy (1 signature).

liran-funaro: I'm updating the test accordingly

protoutil/blockutils_test.go
```diff
@@ -508,6 +513,93 @@ func TestBlockSignatureVerifierByIdentifier(t *testing.T) {
 	require.Equal(t, msppb.NewIdentity("msp3", []byte("identity3")), signatureSet[1].Identity)
 }
 
+func TestBlockSignatureVerifierWithRealPolicy(t *testing.T) {
+	t.Parallel()
+
+	targetPath := t.TempDir()
+	genesisBlock, err := testcrypto.CreateOrExtendConfigBlockWithCrypto(targetPath, &testcrypto.ConfigBlock{
+		OrdererEndpoints: []*types.OrdererEndpoint{
```
HagarMeir: maybe we should use 4 consenters 
make it a "real" setup

protoutil/blockutils_test.go
```diff
@@ -508,6 +513,93 @@ func TestBlockSignatureVerifierByIdentifier(t *testing.T) {
...
+
+			p := testcrypto.BlockPrepareParameters{
+				ConsenterSigners:    signers,
+				ConsenterIDs:        consenterIDs, // Consenter IDs for BFT mode.
+				UseIdentifierHeader: tc.bft,
+			}
+
+			testBlock := &cb.Block{Data: &cb.BlockData{Data: [][]byte{[]byte("tx-1"), []byte("tx-2")}}}
+
+			t.Log("Verifying block with 3/3 valid signatures")
```
HagarMeir: make this 4/4

protoutil/blockutils_test.go
```diff
@@ -508,6 +513,93 @@ func TestBlockSignatureVerifierByIdentifier(t *testing.T) {
...
+			signedBlock := testcrypto.PrepareBlockHeaderAndMetadata(testBlock, p)
+			require.NoError(t, v.Verify(signedBlock.Header, signedBlock.Metadata))
+
+			t.Log("Verifying block with 2/3 valid signatures")
```
HagarMeir: and this 3/4

protoutil/blockutils_test.go
liran-funaro: done

protoutil/blockutils_test.go
liran-funaro: done

protoutil/blockutils_test.go
liran-funaro: done

protoutil/blockutils_test.go
```diff
@@ -508,6 +513,93 @@ func TestBlockSignatureVerifierByIdentifier(t *testing.T) {
 	require.Equal(t, msppb.NewIdentity("msp3", []byte("identity3")), signatureSet[1].Identity)
 }
 
+func TestBlockSignatureVerifierWithRealPolicy(t *testing.T) {
```
tock-ibm, line 517: In order to validate that the BFT BlockValidation policy indeed works as we think it should, I suggest to run this test on multiple cluster sizes, and various sig multiplicities:
4 consenters: 4/4 3/4 pass, less fail
7 consenters 7/7 6/7 5/7 pass, less fail

In addition, try a test where there are enough sigs but some are bad (as opposed to missing), e.g
3/4 good sigs, 1 bad sig (pass)
2/4 good sigs, 1 bad sig (fail)

protoutil/blockutils_test.go
tock-ibm: If you do use less than 4, then we need unanimous agreement: 3/3 must exist

protoutil/blockutils_test.go
```diff
@@ -508,6 +513,94 @@ func TestBlockSignatureVerifierByIdentifier(t *testing.T) {
...
+			}
+		}
+		require.NotNil(t, signers[i])
+	}
+
+	for _, tc := range []struct {
+		name string
+		bft  bool
+	}{
+		{name: "non-BFT with SignatureHeader", bft: false},
```
HagarMeir: no reason to check non-BFT

liran-funaro: This is not really testing non-bft case. It tests signature with SignatureHeader instead of identity (consenter ID)

protoutil/blockutils_test.go
liran-funaro: done

liran-funaro: Now it pass in the CI

tock-ibm (approved):

PR #108, opened 2026-04-27
replace concatenate bytes with asn1 marshal for block signatures

HagarMeir:
https://github.com/hyperledger/fabric-x-common/issues/109

### DIFF
--- a/protoutil/blockutils.go
+++ b/protoutil/blockutils.go
@@ -22,7 +22,6 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/hyperledger/fabric-x-common/api/msppb"
-	"github.com/hyperledger/fabric-x-common/common/util"
 )
 
 var logger = flogging.MustGetLogger("protoutil")
@@ -288,9 +287,10 @@ func (v *BlockSigVerifier) Verify(header *cb.BlockHeader, metadata *cb.BlockMeta
 				"or no signature header provided in block [%d]: %s", header.GetNumber())
 			continue
 		}
+		messageToSign := &MessageToSign{sigHeader, blockHeaderBytes, md.Value}
 		signatureSet = append(signatureSet, &SignedData{
 			Identity:  signerIdentity,
-			Data:      util.ConcatenateBytes(md.Value, sigHeader, blockHeaderBytes),
+			Data:      messageToSign.ASN1MarshalOrPanic(),
 			Signature: metadataSignature.Signature,
 		})
 	}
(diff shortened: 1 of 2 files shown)


protoutil/messagetosign.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package protoutil
+
+import (
+	"encoding/asn1"
+)
+
+// MessageToSign is used when signing a block.
+type MessageToSign struct {
```
liran-funaro: I suggest deciding on a terminology and using it everywhere.
Here it is a `MessageToSign`, in the prepare method it is a `signingPayload`, and in the verify method it is a `msg` and the signature set's data.

This can be defined in `blockutils.go`. No need for a new file for one struct.

protoutil/messagetosign.go
```diff
@@ -0,0 +1,27 @@
...
+	IdentifierHeader     []byte
+	BlockHeader          []byte
+	OrdererBlockMetadata []byte
+}
+
+// Marshal marshals the message to sign or panics on error.
+func (m *MessageToSign) Marshal() []byte {
```
liran-funaro: To comply with other places (see `api/applicationpb/asn1.go`).
```suggestion
func (m *MessageToSign) ASN1Marshal() []byte {
```

liran-funaro (approved):
LGTM. I have a few minor suggestions.

protoutil/blockutils.go
```diff
@@ -396,3 +396,19 @@ func ReadBlockFromFile(blockPath string) (*cb.Block, error) {
...
+
+// MessageToSign is used when signing a block.
+type MessageToSign struct {
+	IdentifierHeader     []byte
+	BlockHeader          []byte
+	OrdererBlockMetadata []byte
+}
+
+// ASN1Marshal marshals the message to sign or panics on error.
+func (m *MessageToSign) ASN1Marshal() []byte {
```
tock-ibm: In Fabric protoutil they usually call it *MarshalOrPanic, so ASN1MarshalOrPanic maybe?

protoutil/blockutils.go
```diff
@@ -396,3 +396,19 @@ func ReadBlockFromFile(blockPath string) (*cb.Block, error) {
...
+	bytes, err := asn1.Marshal(*m)
```
liran-funaro: Lint issue
```suggestion
	rawBytes, err := asn1.Marshal(*m)
```

protoutil/blockutils.go
tock-ibm: The asn1.Marshal() expects a struct value, not a pointer, so the logical thing to do would be to have this method have a value receiver, not a pointer receiver. A value receiver does not change the original struct sent to it.

i.e.
```
// ASN1Marshal marshals the message to sign or panics on error.
func (m MessageToSign) ASN1Marshal() []byte {
	bytes, err := asn1.Marshal(m)
	if err != nil {
		panic(err)
	}
	return bytes
}
```

tock-ibm (approved):

PR #110, opened 2026-04-28
asn1 unmarshal and test and new identifier header

HagarMeir:


protoutil/blockutils_test.go
```diff
@@ -818,3 +818,61 @@ func TestVerifyTransactionsAreWellFormed(t *testing.T) {
...
 }
+
+func TestMessageToSignASN1MarshalUnmarshal(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		msg         protoutil.MessageToSign
+		expectError bool
+	}{
```
liran-funaro: minor: `expectError` is never used.
nit: We usually put the tests struct inline the for `for _, tc := range []struct{} {}`
nit: The test with empty data is redundant. I don't think it adds value.

liran-funaro (approved):
Looks good. Just a minor comment regarding the test.

liran-funaro (approved):

PR #113, opened 2026-05-12
remove unused parameters from config block

### LINKED ISSUE #102: remove irrelevant config parameter from configtx.yaml

cendhu:
#### Type of change

#### Description

Remove legacy Fabric ACLs and no-capability gates

 - drop lifecycle, lscc, and legacy default ACL resources
 - remove sample config capabilities and ACL entries
 - allow Fabric-X configs without capabilities
 - use consensus type instead of capabilities for BFT handling
 - update tests for no-capability Fabric-X baseline
  
#### Related issues

 - resolves #102


liran-funaro (approved):
LGTM. @tock-ibm should also review.
I'm only worried that removing the ACL will remove the only reference we have for the upcoming ACL work. But we can always look it up in Fabric.

cendhu:
> LGTM. @tock-ibm should also review. I'm only worried that removing the ACL will remove the only reference we have for the upcoming ACL work. But we can always look it up in Fabric.

Yes, we can refer the Fabric code. This PR retains the ACL core library and removes only the usage.

common/deliverclient/verifier_assembler.go
```diff
@@ -42,15 +42,11 @@ func (bva *BlockVerifierAssembler) VerifierFromConfig(configuration *common.Conf
 		return createErrorFunc(err), err
 	}
 
-	bftEnabled := bundle.ChannelConfig().Capabilities().ConsensusTypeBFT()
+	cfg, ok := bundle.OrdererConfig()
```
tock-ibm, line 45: check the `ok` and if it is `false` emit an error like in the original code `err := errors.New("no orderer section in config block")`

tools/configtxgen/encoder.go
```diff
@@ -185,9 +185,6 @@ func NewChannelGroup(conf *Profile) (*cb.ConfigGroup, error) {
 //
 //nolint:gocognit // cognitive complexity 23.
 func NewOrdererGroup(conf *Orderer, channelCapabilities map[string]bool) (*cb.ConfigGroup, error) {
-	if conf.OrdererType == "BFT" && !channelCapabilities["V3_0"] {
-		return nil, errors.Errorf("orderer type BFT must be used with V3_0 channel capability: %v", channelCapabilities)
-	}
 	if len(conf.Addresses) > 0 && channelCapabilities["V3_0"] {
 		return nil, errors.Errorf("global orderer endpoints exist, but can not be used with V3_0 capability: %v", conf.Addresses)
 	}
```
tock-ibm: remove this V3_0 check as well, make sure there are no global endpoints regardless

common/deliverclient/verifier_assembler.go
cendhu: Fixed. Added `ok` check — returns error `"no orderer section in config block"` when `OrdererConfig()` returns false. See e4a448f8b.

tools/configtxgen/encoder.go
cendhu: Fixed. Removed V3_0 capability check — global orderer endpoints are now rejected unconditionally. Also removed V3_0 gating in `NewOrdererOrgGroup` — missing org endpoints are now always an error. Kept `channelCapabilities` parameter in both signatures for future reuse. See e4a448f8b.

tock-ibm (approved):

PR #115, opened 2026-05-19
Restore PeerPropose ACL resource and remove ACL capability gates

cendhu:
#### Type of change

- Bug fix

#### Description

Commit c948ab27e removed all ACL resource constants and default policy mappings, but fabric-smart-client's fabricx membership service still depends on:

  - resources.Peer_Propose constant ("peer/Propose")
  - default ACL mapping Peer_Propose → CHANNELWRITERS
  - aclmgmt.ACLProvider / aclmgmt.NewACLProvider

Without these, FSC fails to compile and CheckACL would hit "Unmapped policy" at runtime.

Additionally, remove the capability gate that rejected ACLs when the application capability version was below V1_2. In Fabric-X, ACLs are always allowed regardless of capability version.

Changes:
  - core/aclmgmt/resources/resources.go: restore Peer_Propose constant
  - core/aclmgmt/defaultaclprovider.go: restore Peer_Propose → CHANNELWRITERS mapping
  - common/capabilities/application.go: ACLs() always returns true
  - common/channelconfig/application.go: remove capability validation gate
  - Update tests for always-allowed ACLs

### DIFF
--- a/common/capabilities/application.go
+++ b/common/capabilities/application.go
@@ -69,9 +69,10 @@ func (ap *ApplicationProvider) Type() string {
 	return applicationTypeName
 }
 
-// ACLs returns whether ACLs may be specified in the channel application config
+// ACLs returns whether ACLs may be specified in the channel application config.
+// In Fabric-X, ACLs are always allowed regardless of capability version.
 func (ap *ApplicationProvider) ACLs() bool {
-	return ap.v12 || ap.v13 || ap.v142 || ap.v20 || ap.v25
+	return true
 }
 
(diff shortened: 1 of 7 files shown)


liran-funaro (commented):
We should add it back to the sample YAML file

cendhu:
> We should add it back to the sample YAML file

Done.

liran-funaro (approved):
LGTM. I suggest @mbrandenburger testing it before merging.

mbrandenburger (approved):
Tested it locally and seems to work. Thanks

PR #116, opened 2026-05-26
Fix newest iterator on empty file ledger

cendhu:
#### Type of change

- Bug fix

#### Description

When `SeekPosition_Newest` is used on an empty `FileLedger`, `Iterator` computed `Height - 1`, underflowing to `math.MaxUint64`. This makes the iterator request an impossible block number instead of waiting from block zero.

- Guard newest seek when ledger height is zero
- Start empty-ledger newest iterators at block zero
- Add regression coverage for the empty-ledger case

#### Related

Companion for hyperledger/fabric-x-committer#601.


liran-funaro (approved):
LGTM

liran-funaro (approved):

PR #117, opened 2026-06-03
Add StreamAllTransactions to notification service API

liran-funaro:
#### Type of change

- New feature

#### Description

- Add StreamAllTransactions to notification service API

#### Related issues

- required for https://github.com/hyperledger/fabric-x-committer/issues/618


api/committerpb/notify.proto
```diff
@@ -45,3 +50,49 @@ message RejectedTxIds {
     repeated string tx_ids = 1; // List of rejected transaction IDs.
     string reason = 2;          // The reason for rejection.
 }
+
+// StreamAllRequest is sent by clients to subscribe to all committed transactions.
+message StreamAllRequest {
```
cendhu, line 55: Please remove this and accept an empty message or just keep it empty. Let's keep it simple as we don't know whether they are really needed. Further, it will simplify the review process.

ale-linux: `filter_namespaces` and `include_read_write_sets` is definitely useful for us tho...

liran-funaro: As per the discussion in #618, @ale-linux requested some filtering.
As this only amounts to a small number of code lines, we can keep it.

api/committerpb/notify.proto
```diff
@@ -45,3 +50,49 @@ message RejectedTxIds {
...
+
+// TxEventBatch contains a batch of transaction events from a single block.
+message TxEventBatch {
+    // The block number from which these transactions originated.
+    uint64 block_number = 1;
+
+    // List of transaction events (may be filtered based on request).
+    repeated TxEvent events = 2;
+}
+
```
cendhu, line 81: Maybe we can rename this to `StreamRequest` without `All` as we have filters.

cendhu (approved):
LGTM.

PR #126, opened 2026-06-21
[BREAKING] [dependency] Replace unmaintained library govaluate with expr

### LINKED ISSUE #125: [dependency] unmaintained library `github.com/Knetic/govaluate`
Should be replaced with a maintained fork `github.com/expr-lang/expr`

liran-funaro:
#### Type of change

- Breaking change

#### Description

- Replace unmaintained library `github.com/Knetic/govaluate` with a maintained version `github.com/expr-lang/expr`
- Reduce to a single compile iteration instead of 3 iterations
- Add ID deduplication
- Use case-insensitive gates and roles

#### Related issues

- resolves #125

### DIFF
--- a/common/policydsl/policyparser.go
+++ b/common/policydsl/policyparser.go
@@ -10,10 +10,10 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/expr-lang/expr"
+	"github.com/expr-lang/expr/ast"
 	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
 	mb "github.com/hyperledger/fabric-protos-go-apiv2/msp"
 	"google.golang.org/protobuf/proto"
@@ -26,23 +26,35 @@ type parser struct {
 
 // Gate values.
 const (
-	GateAnd   = "And"
-	GateOr    = "Or"
-	GateOutOf = "OutOf"
+	GateAnd   = "and"
+	GateOr    = "or"
+	GateOutOf = "outof"
 )
 
-var roleRegex = regexp.MustCompile("^([[:alnum:].-]+)[.](admin|member|client|peer|orderer)$")
+// Pattern matches a role case-insensitively:
+//
+//	^                          - Assert start of string
+//	(?i)                       - Enable case-insensitivity
+//	(                          - Group 1 start: Identifier
+//	  [[:alnum:].-]+           - Alphanumeric, dot, or hyphen (1 or more)
+//	)                          - Group 1 end
+//	[.]                        - Match a literal dot
+//	(                          - Group 2 start: Role
+//	  admin|member|client|...  - Specific system role options
+//	)                          - Group 2 end
+//	$                          - Assert end of string
+var roleRegex = regexp.MustCompile("^(?i)([[:alnum:].-]+)[.](admin|member|client|peer|orderer)$")
 
 // FromString takes a string representation of the policy,
 // parses it and returns a SignaturePolicyEnvelope that
 // implements that policy. The supported language is as follows:
 //
-// GATE(P[, P])
-// OutOf(N, P[, P])
+//	GATE(P[, P])
+//	OutOf(N, P[, P])
 //
 // where:
-//   - GATE is either "And" (also "AND" or "and") or "Or" (also "OR" or "or")
-//   - OutOf also supports "OUTOF" or "outof"
+//   - GATE is either "And" or "Or" (case-insensitive)
+//   - OutOf is case-insensitive
 //   - P is either a principal or another nested call to GATE or OutOf
 //   - N is a positive number up to the number of arguments
 //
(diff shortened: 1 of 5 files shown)


common/policydsl/policyparser.go
```diff
@@ -13,341 +13,204 @@ import (
...
 )
 
-// Role values for principals
-const (
-	RoleAdmin   = "admin"
-	RoleMember  = "member"
-	RoleClient  = "client"
-	RolePeer    = "peer"
-	RoleOrderer = "orderer"
-)
```
cendhu, line 36: wondering why this was exported earlier. Maybe client sdk used this in Fabric? Anyway, if we need to expose these constant, we can do so via api/types.go.

common/policydsl/policyparser.go
```diff
@@ -13,341 +13,204 @@ import (
...
+		policy,
+		expr.Function(GateAnd, p.and),
+		expr.Function(strings.ToLower(GateAnd), p.and),
+		expr.Function(strings.ToUpper(GateAnd), p.and),
+		expr.Function(GateOr, p.or),
+		expr.Function(strings.ToLower(GateOr), p.or),
+		expr.Function(strings.ToUpper(GateOr), p.or),
+		expr.Function(GateOutOf, p.nOutOf),
+		expr.Function(strings.ToLower(GateOutOf), p.nOutOf),
+		expr.Function(strings.ToUpper(GateOutOf), p.nOutOf),
```
cendhu: Can we allow only And, AND, and and not aND? This is fine if there is no simple alternative. Anyway, this did not create a problem in fabric so far.

common/policydsl/policyparser.go
```diff
@@ -13,341 +13,204 @@ import (
...
+		p.principals = append(p.principals, principal)
+		p.principalsIDMap[key] = principalID
 	}
 
-	p := &cb.SignaturePolicyEnvelope{
-		Identities: ctx.principals,
-		Version:    0,
-		Rule:       rule,
+	// Create a SignaturePolicy that requires a signature from the principal we've just built.
+	return SignedBy(principalID), nil
```
cendhu, line 189: In the earlier code, we didn't deduplicate principals (there is a TODO). Here, we do. So even if the identities are duplicate, the rule should ask signature for each. A signature is consumed only once and for duplicate identities, signature from distinct members from the org is needed. It is achieved by this line.

common/policydsl/policyparser.go
```diff
@@ -13,341 +13,204 @@ import (
...
+	switch arg := args[0].(type) {
+	case int:
+		t = arg
+	case float64:
+		t = int(arg)
+	case string:
+		var err error
+		t, err = strconv.Atoi(arg)
+		if err != nil {
+			return nil, fmt.Errorf("unrecognized type, expected a number, got %s", reflect.TypeOf(args[0]))
```
cendhu: As compared to the existing code below, the line 105 can reject '1.0'. I am okay with this change. I also wonder why existing code convert 1.9 to 1 which we replicate as well.

```go
// govaluate treats all numbers as float64 only. But and/or may pass int/string. Allowing int/string for flexibility of caller
	if n, ok := arg0.(float64); ok {
		toret += strconv.Itoa(int(n))
	} else if n, ok := arg0.(int); ok {
		toret += strconv.Itoa(n)
	} else if n, ok := arg0.(string); ok {
		toret += n
	} else {
		return nil, fmt.Errorf("unexpected type %s", reflect.TypeOf(arg0))
	}
```

common/policydsl/policyparser.go
liran-funaro: I found no use of these constants in all the repositories under "hyperledger".
In fact, multiple repositories have duplicates of the `policydsl` code with these constants, which are never used outside of this context.

common/policydsl/policyparser.go
liran-funaro: I don't know what was the original author intent, but I guessed that choosing these options was intentional.
But I see no reason to forbid `aND`. I'll check if `expr` supports that.

common/policydsl/policyparser.go
liran-funaro: Is there an action item for me here?

common/policydsl/policyparser.go
liran-funaro: The answer is in the comment you posted here.
`govaluate` didn't support `int`. It converted everything to `float64`. So there was no way to know if the user put `1` or `1.0`.
I suggest changing this behaviour and only support int. Removing string and float.
What do you think?

common/policydsl/policyparser.go
liran-funaro: The same question goes for the roles. Do we require case-sensitive roles (lowercase only)?

liran-funaro: There is a way to make `expr` functions case-insensitive. I can apply it.

common/policydsl/policyparser.go
cendhu: I see. Okay.

common/policydsl/policyparser.go
cendhu: Sounds good.

common/policydsl/policyparser.go
cendhu: No, I just started my observation. If this is correct, no actions are needed.

common/policydsl/policyparser.go
cendhu: Yes. Make sense. Verify once with Angelo to not break their code unnecessarily.

common/policydsl/policyparser.go
liran-funaro: The same question goes for the roles. Do we require case-sensitive roles (lowercase only)?

cendhu: All cases should be allowed.

adecaro:
I'm good with this but I think it introduces a breaking change because we cannot guarantee the two libraries agree on all inputs.

cendhu (approved):
LGTM.

PR #137, opened 2026-07-19
Known certifacates support

gengur:
#### Type of change
improvement to code

#### Description

The fabric-common-x cryptogen generates known certificates that are propagated into the MSPs, which causes failures in the reconfiguration tests.

#### Related issues

https://github.com/hyperledger/fabric-x-orderer/issues/1014

### DIFF
--- a/api/ordererpb/configuration.proto
+++ b/api/ordererpb/configuration.proto
@@ -32,14 +32,16 @@ message PartyConfig {
     repeated bytes CACerts = 2;
     // the certificates of the certificate authorities who generates the party's TLS key-pairs
     repeated bytes TLSCACerts = 3;
+     // the certificates of the known parties' signing certificates that are used to verify signatures
+    repeated bytes KnownCerts = 4;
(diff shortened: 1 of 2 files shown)


api/ordererpb/configuration.proto
```diff
@@ -32,14 +32,16 @@ message PartyConfig {
     repeated bytes CACerts = 2;
     // the certificates of the certificate authorities who generates the party's TLS key-pairs
     repeated bytes TLSCACerts = 3;
+     // the certificates of the known parties' signing certificates that are used to verify signatures
+    repeated bytes KnownCerts = 4;
```
tock-ibm, line 36: this should not be coming from here... only in the known certs in the org definition

tock-ibm (changes requested):

api/ordererpb/configuration.proto
gengur: Right. Closing the PR

PR #138, opened 2026-07-19
[utils] Add gRPC server bootstrapping packages [serve, connection, and retry]

### LINKED ISSUE #675: Discussion: publish `utils/serve` to fabric-x-common
## Context

fabric-x-evm is expected to need the same gRPC server bootstrapping that `utils/serve` provides here (health-check service, TLS provider, interceptors, config) for the endorsement feature. Rather than evm copying the code or depending directly on fabric-x-committer, it can be worthwhile to publish `serve` into fabric-x-common, which fabric-x-committer already depends on.

Advantages:

- Single source of truth - bug/security fixes to server bootstrapping (TLS, retry, health checks) are shared.
- Consistency - evm, committer and future components get the same battle-tested gRPC serving behavior, with similar configuration flags.
- Right dependency direction - extends the existing common ← committer relationship instead of evm depending on committer or hand-copying code.
- Shared maintenance burden - the code is reviewed/owned by more than one team, rather than being committer's problem alone.

It's possible that other (official or unofficial) components will be created later that also need GRPC connectivity and want to fit in the Fabric ecosystem. Moving the code to fabric-x-common might even make that more inviting. As an example, the [custom endorser sample](https://github.com/hyperledger/fabric-x-samples/tree/main/custom-endorser) was easy to build with few lines of code thanks to depending on the committer `utils/serve` package. At the same time you can also see it as an example of a sub-optimal dependency direction...

## What would move

Just three packages, forming a clean, shallow chain: `serve → connection → retry`. `retry` has no internal fabric-x-committer dependencies at all; `connection` depends only on `retry`; `serve`'s production code (`server.go`, `config.go`, `interceptors.go`, `start_serve.go`, `tls_provider.go`) depends only on `connection`. No coordinator/sidecar/VC concepts anywhere in the chain.

## Decoupling check

At first sight, the dependency graph seems to suggest a reliance on committer specific code. However this is only through tests, but not through production code. So some rewriting of tests would be necessary to keep the same coverage, but the scope is limited.

## Alternatives considered (for fabric-x-evm)

- **Copy the code into fabric-x-evm**: avoids the dependency question but risks silent drift — bug fixes and security patches applied in one repo won't reach the other.
- **evm depends directly on fabric-x-committer**: wrong direction, and drags in unrelated coordinator/sidecar/VC concerns that evm has no use for.

## Ask

Discuss whether publishing `serve` (+ `connection` and `retry`) to fabric-x-common is agreeable.

dean-amar:
#### Type of change

- Improvement (improvement to code)

#### Description

- Introduce the gRPC server-bootstrapping chain - `serve` -> `connection` -> `retry`.
In addition, importing the generic `channel` utility and shared `test` helpers, used in the mentioned packages above, from `fabric-x-committer`.

- Hosting them in `fabric-x-common` lets `fabric-x-committer`, `fabric-x-evm`, and
future components share a single implementation instead of duplicating it or relying on `fabric-x-committer` across repos.

Packages added:

- `utils/retry` - backoff/retry.
- `utils/connection` - gRPC client dialing, endpoints, TLS config, health checks.
- `utils/serve`      - gRPC/HTTP server lifecycle, interceptors, TLS provider
- `utils/channel`    - generic channel + readiness primitives (`Ready` latch, buffered reader/writer)

#### Related issues

- Partially resolves https://github.com/hyperledger/fabric-x-committer/issues/675


arner:
Hi @dean-amar , are you still planning to continue this?

@tock-ibm this is what @cendhu mentioned in the call today. Even if you don't want to use the shared package for GRPC, it's good for inspiration ;)

dean-amar:
> Hi @dean-amar , are you still planning to continue this?
> 
> @tock-ibm this is what @cendhu mentioned in the call today. Even if you don't want to use the shared package for GRPC, it's good for inspiration ;)

Hey @arner, this work item is still planned.

Some important features are still being added to these packages. As mentioned above, the migration will be delayed slightly until everything is fully covered and ready to be moved into the common repository.

PR #139, opened 2026-07-20
Add snapshot/checkpoint MALFORMED status codes

cendhu:
#### Type of change

- New feature

#### Description

Add two Status enum values to committerpb for sidecar snapshot/checkpoint form validation:

- MALFORMED_SNAPSHOT_NOT_MARKER_ONLY (116): a _snapshot TX has a non-empty read-write set; only a marker-only _snapshot (entirely empty read-write set) is valid.
- MALFORMED_CHECKPOINT_INVALID_KEY (117): a _checkpoint TX read-write key does not decode as a valid TxHeight.

Both are placed in the stored-in-DB MALFORMED block so they carry a valid TX ID and are recorded to block resubmission. No existing enum value is renumbered. Generated status.pb.go regenerated via make proto.


liran-funaro (approved):
LGTM

PR #145, opened 2026-07-23
Add snapshot rejection status codes

### LINKED ISSUE #143: I-5: [proto] Snapshot transaction rejection status codes
Part of the database snapshot & checkpoint feature (see `specs/001-db-snapshot/rfc-db-snapshot.md` in fabric-x-committer).

## Summary

Add status codes to `api/committerpb/status.proto` so the committer can reject a snapshot transaction that cannot be accepted given the current snapshot lifecycle state, and report a durable outcome for it.

## Added statuses

| Value | Name | Meaning |
|---|---|---|
| 119 | `REJECTED_SNAPSHOT_IN_PROGRESS` | A prior committed snapshot is not yet `CHECKPOINTED`, so a new snapshot cannot be accepted. |
| 120 | `REJECTED_SNAPSHOT_NO_CHECKPOINT` | A prior snapshot exists that was never checkpointed. |
| 121 | `REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK` | More than one snapshot TX appeared in the same block; only the first is processed and the rest are rejected regardless of the first's outcome. |

## Design notes

- All three are **well-formed** transactions with a valid, extractable TX ID. They are placed in the **stored** band (alongside `ABORTED_MVCC_CONFLICT`): the outcome is recorded in the state DB. Storing does not block legitimate retries, which always use a fresh TX ID (resubmitting the same TX ID hits `REJECTED_DUPLICATE_TX_ID`).
- `REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK` (121) is intentionally distinct from `REJECTED_SNAPSHOT_IN_PROGRESS` (119): when two snapshot TXs share a block, the second must be rejected even if the first turns out invalid — reusing `IN_PROGRESS` would falsely claim a snapshot is in flight when none is.

## Scope

- [x] `fabric-x-common`: add the three status codes, regenerate `status.pb.go` via `make proto`.
- [ ] `fabric-x-committer`: enforce the rejection logic (sidecar/coordinator/VC), record the stored status, and align the RFC prose (currently states every snapshot request is accepted / "no REJECTED").

## Downstream impact

Additive enum values only; no renames or renumbering. Backward compatible for `fabric-x-orderer`, `fabric-x-committer`, and `fabric-x`.

cendhu:
#### Type of change

- New feature

#### Description

Add three REJECTED_* status codes for snapshot transaction validation:

- REJECTED_SNAPSHOT_IN_PROGRESS (119): a prior committed snapshot is not yet CHECKPOINTED, so a new snapshot cannot be accepted.
- REJECTED_SNAPSHOT_NO_CHECKPOINT (120): a prior snapshot exists that was never checkpointed.
- REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK (121): more than one snapshot TX appeared in the same block; only the first is processed and the rest are rejected regardless of the first's outcome.

All three are well-formed transactions with a valid, extractable TX ID, so they are placed in the stored band (like ABORTED_MVCC_CONFLICT): the outcome is recorded in the state database. Storing does not block legitimate retries, which always use a fresh TX ID.

Regenerated status.pb.go via make proto.

#### Related issues

  - resolves #143

### DIFF
--- a/api/committerpb/status.proto
+++ b/api/committerpb/status.proto
@@ -58,4 +58,7 @@ enum Status {
     MALFORMED_SNAPSHOT_NOT_MARKER_ONLY = 116;           // _snapshot TX has a non-empty read-write set; only a marker-only _snapshot (entirely empty read-write set) is valid.
     MALFORMED_CHECKPOINT_INVALID_KEY = 117;             // _checkpoint TX read-write key does not decode as a valid TxHeight.
     MALFORMED_SYSTEM_TX_NOT_STANDALONE = 118;           // System TX must be standalone and cannot be mixed with other namespaces.
+    REJECTED_SNAPSHOT_IN_PROGRESS = 119;                // Another snapshot is still being hashed / awaiting its checkpoint (a prior _snapshot row is not yet CHECKPOINTED).
+    REJECTED_SNAPSHOT_NO_CHECKPOINT = 120;              // A prior snapshot exists that was never checkpointed, so a new snapshot cannot be accepted.
+    REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK = 121;         // More than one snapshot TX appeared in the same block; only the first is processed and the rest are rejected regardless of the first's outcome.
 }
--- a/api/committerpb/status.pb.go
+++ b/api/committerpb/status.pb.go
@@ -60,6 +60,9 @@ const (
 	Status_MALFORMED_SNAPSHOT_NOT_MARKER_ONLY        Status = 116 // _snapshot TX has a non-empty read-write set; only a marker-only _snapshot (entirely empty read-write set) is valid.
 	Status_MALFORMED_CHECKPOINT_INVALID_KEY          Status = 117 // _checkpoint TX read-write key does not decode as a valid TxHeight.
 	Status_MALFORMED_SYSTEM_TX_NOT_STANDALONE        Status = 118 // System TX must be standalone and cannot be mixed with other namespaces.
+	Status_REJECTED_SNAPSHOT_IN_PROGRESS             Status = 119 // Another snapshot is still being hashed / awaiting its checkpoint (a prior _snapshot row is not yet CHECKPOINTED).
+	Status_REJECTED_SNAPSHOT_NO_CHECKPOINT           Status = 120 // A prior snapshot exists that was never checkpointed, so a new snapshot cannot be accepted.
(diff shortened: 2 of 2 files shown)


api/committerpb/status.proto
```diff
@@ -1,61 +1,61 @@
...
 message TxStatus {
-    TxRef ref = 1;
-    Status status = 2;
+  TxRef ref = 1;
+  Status status = 2;
 }
 
-// Status represents the result of transaction validation.
-// Except for STATUS_UNSPECIFIED, all statuses are recorded in the ledger.
-// Some statuses are also stored in the state database which prevent resubmission of the same transaction ID.
```
liran-funaro, line 28: Why did you removed this comment?

api/committerpb/status.proto
```diff
@@ -1,61 +1,61 @@
 /*
-Copyright IBM Corp. All Rights Reserved.
+   Copyright IBM Corp. All Rights Reserved.
 
-SPDX-License-Identifier: Apache-2.0
+   SPDX-License-Identifier: Apache-2.0
```
liran-funaro: We didn't add this indentation in other files.

api/committerpb/status.proto
```diff
@@ -1,61 +1,61 @@
...
 
 syntax = "proto3";
 
-option go_package = "github.com/hyperledger/fabric-x-common/api/committerpb";
-
 package committerpb;
 
 import "api/committerpb/ref.proto";
 
+option go_package = "github.com/hyperledger/fabric-x-common/api/committerpb";
```
liran-funaro: Why was it moved here?

liran-funaro (commented):
LGTM, but I don't understand the format changes.

api/committerpb/status.proto
cendhu: This happened due to the mix of nvim (manual edits).

api/committerpb/status.proto
cendhu: This happened due to the mix of nvim (manual edits).

api/committerpb/status.proto
cendhu: Fixed: restored original proto style/order/comments. Diff now only adds the three snapshot status values plus generated status.pb.go updates.

api/committerpb/status.proto
cendhu: Fixed: restored original proto style/order/comments. Diff now only adds the three snapshot status values plus generated status.pb.go updates.

api/committerpb/status.proto
cendhu: Fixed: restored original proto style/order/comments. Diff now only adds the three snapshot status values plus generated status.pb.go updates.

liran-funaro (approved):

PR #147, opened 2026-08-09
fxadmin cli skeleton

MayRosenbaum:
#### Type of change
Admin CLI tool for reconfiguration

#### Description
Skeleton of the admin CLI tool

### DIFF
--- a/tools/fxadmin/core/cli/cli.go
+++ b/tools/fxadmin/core/cli/cli.go
@@ -1,0 +1,214 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package cli builds the fxadmin command tree. fxadmin administers the
+// configuration of Fabric-X: it pulls the current configuration from the
+// running network, decodes it, computes and endorses a configuration update,
+// wraps it into a submittable transaction, submits it, and follows the
+// assembler ledger until the change is committed. fxadmin operates on the
+// single channel "arma".
+package cli
+
+import (
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/cockroachdb/errors"
+)
+
+// Channel is the single, fixed channel fxadmin operates on.
+const Channel = "arma"
+
+// Flag names shared across commands.
+const (
+	flagConfig       = "config"
+	flagCurrentBlock = "current-block"
+	flagOutput       = "output"
+	flagBlock        = "block"
+	flagTimeout      = "timeout"
+)
+
+// CLI is the fxadmin command-line application. It owns the kingpin command
+// tree and a dispatch table mapping each command to a closure
+// that invokes the corresponding Handlers method with the bound values.
+type CLI struct {
+	app      *kingpin.Application
+	handlers Handlers
+	dispatch map[string]func() error
+}
+
+// New builds the fxadmin CLI around the given handlers. The version string is
+// surfaced through the --version flag. No command is executed until Run is called.
+func New(handlers Handlers, version string) *CLI {
+	app := kingpin.New("fxadmin", "Fabric-X reconfiguration admin CLI.")
+	app.Version(version)
+
+	c := &CLI{
+		app:      app,
+		handlers: handlers,
+		dispatch: make(map[string]func() error),
+	}
+
+	c.addLedgerCommands()
+	c.addDecodeCommand()
+	c.addComputeUpdateCommand()
+	c.addTxCommands()
+	c.addFollowCommand()
+
+	return c
+}
+
+// Parse resolves args to the selected command and binds flag/argument values.
+func (c *CLI) Parse(args []string) (string, error) {
+	cmd, err := c.app.Parse(args)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse arguments")
+	}
+	return cmd, nil
+}
+
+// Run parses args and executes the selected command via its handler.
+func (c *CLI) Run(args []string) error {
+	cmd, err := c.Parse(args)
+	if err != nil {
+		return err
+	}
+	run, ok := c.dispatch[cmd]
+	if !ok {
+		return errors.Newf("no handler registered for command %q", cmd)
+	}
+	return run()
+}
+
+// register records the closure that executes cmd after a successful parse.
+func (c *CLI) register(cmd *kingpin.CmdClause, run func() error) {
+	c.dispatch[cmd.FullCommand()] = run
+}
+
+// addLedgerCommands wires `fxadmin ledger` and its height/block/config
+// subcommands. --config and --current-block are common to all of them.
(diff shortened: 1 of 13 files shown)


tools/fxadmin/core/cli/cli.go
```diff
@@ -0,0 +1,238 @@
...
+// single channel "arma".
+package cli
+
+import (
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/cockroachdb/errors"
+)
+
+// Channel is the single, fixed channel fxadmin operates on.
+const Channel = "arma"
```
tock-ibm, line 21: The channel name is not constant - it is taken from the config block.

tools/fxadmin/core/cli/cli.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// Parse resolves args to the selected command and binds flag/argument values.
+func (c *CLI) Parse(args []string) (string, error) {
+	cmd, err := c.app.Parse(args)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse arguments")
+	}
+	return cmd, nil
+}
```
tock-ibm, line 69: 🟡 **Minor:** `Parse` is exported, but `Run` (same file) is its only caller and there's no external consumer yet (no `cmd/fxadmin/main.go` in this PR). Per [AGENTS.md](AGENTS.md) exported symbols are a cross-repo contract, so a two-phase public parse/run entry point commits us to supporting it before a caller demonstrates the need.

**Recommendation:** unexport until a caller needs it — the tests only call `Run`, so nothing breaks.

```suggestion
// parse resolves args to the selected command and binds flag/argument values.
func (c *CLI) parse(args []string) (string, error) {
```
(and update the `c.Parse(args)` call in `Run` to `c.parse(args)`.)

tools/fxadmin/core/cli/cli.go
```diff
@@ -0,0 +1,238 @@
...
+)
+
+// Channel is the single, fixed channel fxadmin operates on.
+const Channel = "arma"
+
+// Flag names shared across commands.
+const (
+	flagConfig       = "config"
+	flagCurrentBlock = "current-block"
+	flagOutput       = "output"
```
tock-ibm, line 27: 🔵 **Nit:** `Channel` is exported but unread anywhere in the PR (the handlers are all skeletons). Because it's exported, the `unused` linter won't flag it, so it can silently drift from the real channel name until a later PR wires it in. Fine to keep as documented intent for a skeleton — just flagging that it's a cross-repo-visible symbol. Consider unexporting it (`channel`) until it's actually consumed.

tools/fxadmin/core/tx/tx.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Package tx implements the `fxadmin tx` command family: endorsing, merging,
+// preparing, submitting, and sending configuration update transactions.
+package tx
+
+import (
+	"fmt"
+
+	"github.com/cockroachdb/errors"
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
```
tock-ibm, line 15: 🔵 **Nit:** the `errNotImplemented` sentinel + `Handler` + `New()` + `fmt.Errorf("...: %w", errNotImplemented)` shape is copy-pasted verbatim across `tx`, `decode`, `follow`, `ledger`, and `update`. `dupl` is enabled (default settings) and *may* flag these structurally identical files.

Largely inherent to a skeleton and the handlers diverge as soon as they're implemented, so no shared abstraction is warranted — but if `dupl` fires in CI, a scoped `//nolint:dupl // skeleton placeholder` (with the reason) resolves it. (Same pattern at `decode.go`, `follow.go`, `ledger.go`, `compute_update.go`.)

tools/fxadmin/core/decode/decode_test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/decode"
+)
+
+// TestRunNotImplemented asserts the decode command is a skeleton that reports
+// "not implemented". Replace with a proto-to-JSON round-trip test once the
```
tock-ibm, line 18: 🔵 **Nit:** `TestRunNotImplemented` here is near-identical to the ones in `follow_test.go` and `compute_update_test.go` (differ only in constructor + args). Fine for a skeleton — the `// Replace with ...` comments already flag these as placeholders — noting only so it's a conscious choice.

tock-ibm (commented):
## Review Summary — fxadmin cli skeleton

Thanks for this — it's a clean, well-scoped skeleton. We introduce the `fxadmin` command tree (kingpin + a dispatch map) and five `core/*` handler packages that currently return `not implemented`, with tests that genuinely exercise the real command tree.

**Conventions ✅** — Apache-2.0 headers on all files; `cockroachdb/errors` at the origin with `fmt.Errorf(...%w...)` for context; one `flogging` logger per package; `New()` returns concrete `*Handler` (no `ireturn`); tests are table-driven with `t.Parallel()` / `require.ErrorContains` / `t.Helper`, no new Ginkgo. `go build` and `go vet` pass, no line exceeds 120 cols, and `kingpin` is already a repo dependency.

**No correctness bugs found** — the dispatch map keyed on `FullCommand()`, the closure flag-pointer capture, and the parse-then-dispatch flow are all correct. `cli_test.go` routing/parse-error coverage is a nice touch (handlers left nil to prove parse failures never reach them).

All findings below are **Minor/Nit** cleanup items — none block.

_Posted as a comment-only review._

tools/fxadmin/core/cli/cli.go
MayRosenbaum: removed channel

tools/fxadmin/core/cli/cli.go
```diff
@@ -0,0 +1,238 @@
...
+	cmd, err := c.Parse(args)
+	if err != nil {
+		return err
+	}
+	run, ok := c.dispatch[cmd]
+	if !ok {
+		return errors.Newf("no handler registered for command %q", cmd)
+	}
+	return run()
+}
```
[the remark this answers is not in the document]
MayRosenbaum: I added a check + test case

tock-ibm (approved):
Fix the comments on this one in the next PRs please.

tools/fxadmin/core/tx/tx.go
MayRosenbaum: dupl does not fire.
Anyway, the `errNotImplemented` will be removed anyway once the commands are implemented.

tools/fxadmin/core/decode/decode_test.go
MayRosenbaum: It's a conscious choice.

PR #149, opened 2026-08-11
Fxadmin ledger commands

MayRosenbaum:
#### Type of change
fxadmin ledger command with implementation

### DIFF
--- a/tools/fxadmin/core/client/client.go
+++ b/tools/fxadmin/core/client/client.go
@@ -1,0 +1,256 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package client connects to the Fabric-X orderer. It combines the user
+// identity and client-TLS material (user.Config) with the orderer
+// connection information read from a config block (ordererconn.Info)
+// to sign and send seek and broadcast envelopes to assemblers and routers.
+package client
+
+import (
+	"context"
+	"encoding/pem"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/cockroachdb/errors"
+	"github.com/hyperledger/fabric-lib-go/bccsp"
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
+	ab "github.com/hyperledger/fabric-protos-go-apiv2/orderer"
+
+	"github.com/hyperledger/fabric-x-common/common/util"
+	"github.com/hyperledger/fabric-x-common/msp"
+	"github.com/hyperledger/fabric-x-common/protoutil"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/ordererconn"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/signer"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/user"
+	"github.com/hyperledger/fabric-x-common/tools/pkg/comm"
+)
+
+const rpcTimeout = 30 * time.Second
+
+var logger = flogging.MustGetLogger("fxadmin.client")
+
+// Client sends signed seek and broadcast envelopes to the orderer
+// using the user identity for signing and TLS.
+type Client struct {
+	ordererConnInfo *ordererconn.Info
+	clientConfig    comm.ClientConfig
+	signer          msp.SigningIdentity
+	tlsCertHash     []byte
+}
+
+// Load builds a Client from the user configuration and the config block.
+func Load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error) {
+	ordererConnInfo, err := ordererconn.Load(block, csp)
+	if err != nil {
+		return nil, err
+	}
+	clientConfig, err := newClientConfig(config, ordererConnInfo)
+	if err != nil {
+		return nil, err
+	}
+	signingIdentity, err := signer.New(config.MSP.LocalMspID, config.MSP.LocalMspDir)
+	if err != nil {
+		return nil, err
+	}
+	tlsCertHash, err := tlsCertHash(clientConfig)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{
+		ordererConnInfo: ordererConnInfo,
+		clientConfig:    clientConfig,
+		signer:          signingIdentity,
+		tlsCertHash:     tlsCertHash,
+	}, nil
+}
+
+// ChannelID returns the channel the client targets.
+func (c *Client) ChannelID() string {
+	return c.ordererConnInfo.ChannelID
+}
+
+// FetchBlock seeks the block described by seek from the first reachable
+// assembler, failing over to the next assembler on error.
+func (c *Client) FetchBlock(seek *ab.SeekInfo) (*cb.Block, error) {
+	envelope, err := c.createSignedDeliverSeekEnvelope(seek)
+	if err != nil {
+		return nil, err
+	}
+
+	var errs []error
+	for _, endpoint := range c.ordererConnInfo.AssemblerEndpoints {
+		block, err := c.fetchBlockFromEndpoint(endpoint, envelope)
+		if err == nil {
+			return block, nil
+		}
+		logger.Debugf("failed to fetch block from assembler %s: %v", endpoint, err)
+		errs = append(errs, fmt.Errorf("assembler %s: %w", endpoint, err))
+	}
+	return nil, errors.Wrap(errors.Join(errs...), "failed to fetch block from all assemblers")
+}
+
+// BroadcastToAllRouters sends a prepared envelope to every router, returning an
+// error if any router rejects or is unreachable. The envelope must already be
+// built and signed by the caller (e.g. a prepared configuration transaction).
+func (c *Client) BroadcastToAllRouters(envelope *cb.Envelope) error {
+	var errs []error
+	for _, endpoint := range c.ordererConnInfo.RouterEndpoints {
+		if err := c.broadcastToRouter(endpoint, envelope); err != nil {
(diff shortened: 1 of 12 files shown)


tools/fxadmin/core/client/operations.go
```diff
@@ -0,0 +1,143 @@
...
+	ab "github.com/hyperledger/fabric-protos-go-apiv2/orderer"
+
+	"github.com/hyperledger/fabric-x-common/protoutil"
+)
+
+const rpcTimeout = 30 * time.Second
+
+// FetchBlock seeks the block described by seek from the first reachable
+// assembler, failing over to the next assembler on error.
+func (c *Client) FetchBlock(seek *ab.SeekInfo) (*cb.Block, error) {
```
tock-ibm: It is a bad idea and anti pattern to have the acceptor methods of a struct `Client` in two different files.

tools/fxadmin/core/ordererconn/ordererconn.go
```diff
@@ -0,0 +1,80 @@
...
+
+	ordererConfig, exists := bundle.OrdererConfig()
+	if !exists {
+		return nil, errors.New("config block has no orderer configuration")
+	}
+
+	var sharedConfig ordererpb.SharedConfig
+	if err := proto.Unmarshal(ordererConfig.ConsensusMetadata(), &sharedConfig); err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal ARMA shared config from consensus metadata")
+	}
```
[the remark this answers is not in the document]
MayRosenbaum: done for safety

tools/fxadmin/core/cli/cli.go
```diff
@@ -60,7 +56,7 @@ func New(handlers Handlers, version string) *CLI {
 }
 
 // Parse resolves args to the selected command and binds flag/argument values.
-func (c *CLI) Parse(args []string) (string, error) {
+func (c *CLI) parse(args []string) (string, error) {
```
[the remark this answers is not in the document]
MayRosenbaum: resolved in prev PR

tools/fxadmin/core/client/operations.go
MayRosenbaum: All code in operations.go is forwarded to client.go

tools/fxadmin/core/ledger/ledger.go
```diff
@@ -5,47 +5,145 @@ SPDX-License-Identifier: Apache-2.0
...
 }
 
-// Height implements `fxadmin ledger height`.
-func (*Handler) Height(configPath, currentBlockPath string) error {
+// Height implements `fxadmin ledger height`. It prints the ledger height (the
+// newest block number plus one).
+func (h *Handler) Height(configPath, currentBlockPath string) error {
 	logger.Debugf("ledger height: config=%s current-block=%s", configPath, currentBlockPath)
-	return fmt.Errorf("ledger height: %w", errNotImplemented)
+	cl, err := h.newOrdererClient(configPath, currentBlockPath)
```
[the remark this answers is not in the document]
MayRosenbaum: will be added next PR

tools/fxadmin/core/client/client.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	resp, err := stream.Recv()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to receive deliver response")
+	}
+
+	block := resp.GetBlock()
+	if block == nil {
+		return nil, errors.Newf("received a non-block deliver response: %T", resp.GetType())
```
tock-ibm, line 160: 🟡 **Minor** — this drops the Deliver `Status`.

When an assembler rejects the seek it replies with a normal `DeliverResponse{Status: FORBIDDEN}` (or `BAD_REQUEST` / `SERVICE_UNAVAILABLE`) — a message on the stream, not a stream error. `resp.GetBlock()` is nil, so we return `"received a non-block deliver response: *orderer.DeliverResponse_Status"` and lose the actual status. Because `FetchBlock` then fails over across every assembler, a channel-policy denial surfaces as a confusing aggregated error that never mentions `FORBIDDEN`.

`broadcastToRouter` already inspects `resp.GetStatus()` correctly — we should do the same here (switch on `resp.GetType()` and surface the status when it's a `*ab.DeliverResponse_Status`).

_AI-generated finding — please verify._

tools/fxadmin/core/client/client.go
```diff
@@ -0,0 +1,256 @@
...
+	var errs []error
+	for _, endpoint := range c.ordererConnInfo.AssemblerEndpoints {
+		block, err := c.fetchBlockFromEndpoint(endpoint, envelope)
+		if err == nil {
+			return block, nil
+		}
+		logger.Debugf("failed to fetch block from assembler %s: %v", endpoint, err)
+		errs = append(errs, fmt.Errorf("assembler %s: %w", endpoint, err))
+	}
+	return nil, errors.Wrap(errors.Join(errs...), "failed to fetch block from all assemblers")
```
tock-ibm, line 96: 🟡 **Minor** — `errors.Wrap(errors.Join(errs...), …)` collapses to `nil` when `AssemblerEndpoints` is empty.

The loop never runs → `errs` is empty → `errors.Join()` returns `nil` → `errors.Wrap(nil, …)` returns `nil`. So `FetchBlock` yields `(nil, nil)`, and callers treat the nil block as valid: `Height` computes `GetHeader().GetNumber()+1` (nil-safe getters → 0) and prints `1`; `Block`/`Config` call `writeBlock(nil, …)` and silently emit a zero block.

Same shape at line 109 for `BroadcastToAllRouters` — it returns `nil` (success) when `RouterEndpoints` is empty, having sent nothing. Today this is only guarded by `ordererconn.Load`'s non-empty checks; these public methods have no guard of their own. Consider an explicit empty-list guard (or asserting the invariant) so an empty endpoint set is a hard error rather than a silent no-op.

_AI-generated finding — please verify._

tools/fxadmin/core/client/client.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if block.GetData() == nil || len(block.GetData().GetData()) == 0 {
+		return nil, errors.New("received an empty block")
+	}
+	return block, nil
+}
+
+// broadcastToRouter opens a Broadcast stream to endpoint, sends the envelope, and
+// checks the response status.
+func (c *Client) broadcastToRouter(endpoint string, envelope *cb.Envelope) error {
```
tock-ibm, line 170: 🟡 **Minor** — `broadcastToRouter` and `fetchBlockFromEndpoint` (line 133) duplicate the entire connection scaffolding.

The two methods differ only in Deliver-vs-Broadcast and block-vs-status handling; the `Dial`, `context.WithTimeout`, `CloseSend`/`Close` defers, and `Send` error wrapping are copy-pasted. This is what the `dupl` linter targets, and `make lint` runs `golangci-lint --new-from-rev=main` over this new code. Extracting the shared dial/stream lifecycle (e.g. a helper that opens the stream and defers cleanup, taking a per-RPC callback) would halve the maintenance surface.

_AI-generated finding — please verify._

tools/fxadmin/core/ordererconn/ordererconn.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build channel config bundle from config block")
+	}
+
+	ordererConfig, exists := bundle.OrdererConfig()
+	if !exists {
+		return nil, errors.New("config block has no orderer configuration")
+	}
+
+	if ordererConfig.ConsensusType() != "arma" {
```
tock-ibm, line 55: 🟡 **Minor** — the `"arma"` consensus type is hardcoded here (and again in the error message on the next line) instead of reusing the existing exported constant.

`tools/configtxgen` already defines `ConsensusTypeArma = "arma"` (`encoder.go:45`) — the canonical name used when the block is built. Duplicating the raw literal means a future rename of the consensus type would leave this reader silently rejecting valid ARMA blocks with `"unsupported consensus type"`. Suggest importing and comparing against `configtxgen.ConsensusTypeArma`.

_AI-generated finding — please verify._

tools/fxadmin/core/client/client.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	clientConfig, err := newClientConfig(config, ordererConnInfo)
+	if err != nil {
+		return nil, err
+	}
+	signingIdentity, err := signer.New(config.MSP.LocalMspID, config.MSP.LocalMspDir)
+	if err != nil {
+		return nil, err
+	}
+	tlsCertHash, err := tlsCertHash(clientConfig)
```
tock-ibm, line 62: 🔵 **Nit** — the local variable `tlsCertHash` shadows the package-level `tlsCertHash` function (defined ~line 238) on the same line.

`tlsCertHash, err := tlsCertHash(clientConfig)` compiles (the RHS resolves the function before the new binding takes scope), but the identifier now names both a function and a value in the same block — a readability trap that would break a second call site added later. Suggest renaming the local (e.g. `certHash`, updating the struct literal at line 70).

_AI-generated finding — please verify._

tock-ibm (commented):
🤖 **AI-generated review** — comment-only, no approval implied. Analyzed the new `fxadmin` ledger/client/ordererconn code against PR head `4275c4089` (builds/vets/tests clean). 5 findings below, all Minor/Nit; nothing blocking. Please verify each independently.

tock-ibm (approved):
LGTM. @MayRosenbaum please fix these minor comments  in a separate PR.

tools/fxadmin/core/client/client.go
MayRosenbaum: over complicated

tools/fxadmin/core/ordererconn/ordererconn.go
MayRosenbaum: done

PR #150, opened 2026-08-11
Added `toString()` and `toDisplayString()` functions to render MSP policy as string

pasquale95:
#### Type of change

- New feature

#### Description

Add `ToString`/`ToDisplayString` to `common/policydsl`, the exact inverse of `FromString`: they render a `*cb.SignaturePolicyEnvelope` back into the policy DSL string it was parsed from (e.g. `OR('Org1MSP.member', 'Org2MSP.member')`).

No renderer existed anywhere in this codebase or its consumers — `fxconfig` even has a commented-out `parsePolicy` with `// TODO: some pretty print would be beautiful`. Consumers that store a `SignaturePolicyEnvelope` (e.g. an MSP-based endorsement policy) have no way to show it back to a user except as raw, unreadable protobuf bytes.

- `ToString` returns an error if any part of the policy has no DSL representation (a bare `IDENTITY` principal, an out-of-range `signed_by`, an unsupported principal classification).
- `ToDisplayString` never fails: it substitutes a readable placeholder (e.g. `<identity:0>`, `<principal:a1b2c3d4>`) for anything unrenderable instead of erroring.

Both share one recursive tree walk parameterized by a strict/lenient mode, so there is a single implementation of the rendering logic behind the two entry points.

### DIFF
--- a/common/policydsl/policyrenderer.go
+++ b/common/policydsl/policyrenderer.go
@@ -7,7 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package policydsl
 
 import (
-	"encoding/hex"
+	"encoding/base64"
 	"fmt"
 	"strings"
 
@@ -16,19 +16,21 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// renderMode selects how ToString and ToDisplayString handle a construct that has no DSL
-// representation: strictMode fails the render, lenientMode substitutes a placeholder.
-type renderMode int
+// principalPlaceholderLen bounds how many bytes of a principal are encoded into a placeholder:
+// an IDENTITY principal carries a full certificate, so dumping it untruncated would put a
+// couple of KB into the rendered string.
+const principalPlaceholderLen = 8
 
+// Gate*Display are the canonical, human-readable gate spellings used by ToString and
+// ToDisplayString. They deliberately do not reuse GateAnd/GateOr/GateOutOf, which are the
+// lowercase parser tokens ("and"/"or"/"outof") rather than the display spelling; Visit
+// lowercases gate identifiers during AST traversal, so this casing still round-trips through
+// FromString.
 const (
-	strictMode renderMode = iota
-	lenientMode
+	GateAndDisplay   = "AND"
+	GateOrDisplay    = "OR"
+	GateOutOfDisplay = "OutOf"
 )
-
-// principalPlaceholderHexLen bounds how many bytes of a principal are hex-dumped into a
-// placeholder: an IDENTITY principal carries a full certificate, so dumping it untruncated
-// would put a couple of KB of hex into the rendered string.
-const principalPlaceholderHexLen = 8
 
 // ToString renders an envelope as a policy DSL expression accepted by FromString. It returns
 // an error if any part of the policy has no DSL representation.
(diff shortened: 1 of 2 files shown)


iliecirciumaru (approved):

common/policydsl/policyrenderer.go
```diff
@@ -0,0 +1,228 @@
...
+	return renderRule(env.GetRule(), env.GetIdentities(), strictMode)
+}
+
+// ToDisplayString renders an envelope for human consumption, substituting readable
+// placeholders for anything with no DSL representation. It never fails.
+//
+// Example: an envelope built with Envelope(SignedBy(0), [][]byte{identityBytes}) — a single
+// IDENTITY principal, which has no DSL syntax — renders to "<identity:0>" instead of erroring.
+func ToDisplayString(env *cb.SignaturePolicyEnvelope) string {
+	rendered, err := renderRule(env.GetRule(), env.GetIdentities(), lenientMode)
```
liran-funaro: Instead of passing mode everywhere (a control flag, which is unfavorable), you can make all methods return the display string and a typed error saying the result isn't parsable.
The ToString will return only the error if exists, and ToDisplayString will dismiss such errors, returning only the string.

common/policydsl/policyrenderer.go
```diff
@@ -0,0 +1,228 @@
...
+// Gate*Display are the canonical, human-readable gate spellings used by ToString and
+// ToDisplayString. They deliberately do not reuse GateAnd/GateOr/GateOutOf, which are the
+// lowercase parser tokens ("and"/"or"/"outof") rather than the display spelling; Visit
+// lowercases gate identifiers during AST traversal, so this casing still round-trips through
+// FromString.
+const (
+	GateAndDisplay   = "AND"
+	GateOrDisplay    = "OR"
+	GateOutOfDisplay = "OutOf"
+)
```
liran-funaro: const should be on top

common/policydsl/policyrenderer.go
```diff
@@ -0,0 +1,228 @@
...
+	return fmt.Sprintf("'%s.%s'", ou.GetMspIdentifier(), ou.GetOrganizationalUnitIdentifier())
+}
+
+// placeholderPrincipal renders raw principal bytes as a hex placeholder, truncated to
+// principalPlaceholderHexLen bytes, e.g. []byte("not an MSPRole") -> "<principal:6e6f7420616e204d>".
+// Truncation matters for IDENTITY principals, which carry a full certificate: dumping it
+// untruncated would put a couple of KB of hex into the rendered string.
+func placeholderPrincipal(raw []byte) string {
+	n := min(len(raw), principalPlaceholderHexLen)
+	return "<principal:" + hex.EncodeToString(raw[:n]) + ">"
```
liran-funaro: minor: I suggest using base64 for a more compact endocing.

common/policydsl/policyrenderer.go
```diff
@@ -0,0 +1,228 @@
...
+	switch t := rule.GetType().(type) {
+	case *cb.SignaturePolicy_NOutOf_:
+		return renderNOutOf(t.NOutOf, identities, mode)
+	case *cb.SignaturePolicy_SignedBy:
+		return renderSignedBy(t.SignedBy, identities, mode)
+	default:
+		if mode == strictMode {
+			return "", fmt.Errorf("policy rule has no DSL representation: %T", t)
+		}
+		return "<unknown-rule>", nil
```
liran-funaro: Is this reachable? I don't see any test for this.

common/policydsl/policyrenderer.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	switch {
+	case n == 1:
+		return fmt.Sprintf("%s(%s)", GateOrDisplay, strings.Join(rendered, ", ")), nil
+	case int(n) == len(rules):
+		return fmt.Sprintf("%s(%s)", GateAndDisplay, strings.Join(rendered, ", ")), nil
+	default:
+		return fmt.Sprintf("%s(%d, %s)", GateOutOfDisplay, n, strings.Join(rendered, ", ")), nil
+	}
```
liran-funaro: nit
```suggestion
	switch int(n) {
	case 1:
		return fmt.Sprintf("%s(%s)", GateOrDisplay, strings.Join(rendered, ", ")), nil
	case len(rules):
		return fmt.Sprintf("%s(%s)", GateAndDisplay, strings.Join(rendered, ", ")), nil
	default:
		return fmt.Sprintf("%s(%d, %s)", GateOutOfDisplay, n, strings.Join(rendered, ", ")), nil
	}
```

liran-funaro (commented):
I did a quick review. I'll do a deep dive after these comments are addressed.

common/policydsl/policyrenderer.go
pasquale95: Resolved.

common/policydsl/policyrenderer.go
pasquale95: Resolved

common/policydsl/policyrenderer.go
pasquale95: Added test

common/policydsl/policyrenderer.go
pasquale95: resolved

common/policydsl/policyrenderer.go
pasquale95: Resolved

liran-funaro (approved):
LGTM

PR #153, opened 2026-08-16
[fxadmin] compute update command

MayRosenbaum:
#### Type of change

- New feature

#### Description

Compute update command in fxadmin.

#### Related issues

### DIFF
--- a/tools/fxadmin/core/decode/decode.go
+++ b/tools/fxadmin/core/decode/decode.go
@@ -13,7 +13,6 @@ import (
 	"bytes"
 	"io"
 	"os"
-	"path/filepath"
 	"reflect"
 
 	"github.com/cockroachdb/errors"
@@ -23,6 +22,7 @@ import (
 	"google.golang.org/protobuf/reflect/protoregistry"
 
 	"github.com/hyperledger/fabric-x-common/protolator"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/utils"
 )
 
 var logger = flogging.MustGetLogger("fxadmin.decode")
@@ -44,7 +44,7 @@ func New() *Handler {
 func (*Handler) Run(blockPath, outputPath string) error {
 	logger.Debugf("decode: block=%s output=%s", blockPath, outputPath)
 
-	if err := requireDistinctPaths(blockPath, outputPath); err != nil {
+	if err := utils.RequireDistinctOutput(outputPath, blockPath); err != nil {
 		return err
 	}
 
@@ -64,23 +64,6 @@ func (*Handler) Run(blockPath, outputPath string) error {
 	}
 
 	logger.Infof("decoded %s to %s\n", blockPath, outputPath)
-	return nil
-}
-
-// requireDistinctPaths rejects a block and output that resolve to the same
-// file, so decoding never overwrites its own source.
-func requireDistinctPaths(blockPath, outputPath string) error {
-	blockAbs, err := filepath.Abs(blockPath)
-	if err != nil {
-		return errors.Wrapf(err, "failed to resolve block path %q", blockPath)
-	}
-	outputAbs, err := filepath.Abs(outputPath)
-	if err != nil {
-		return errors.Wrapf(err, "failed to resolve output path %q", outputPath)
-	}
-	if blockAbs == outputAbs {
-		return errors.Newf("block and output must be different files, both resolve to %q", blockAbs)
-	}
 	return nil
 }
 
--- a/tools/fxadmin/core/decode/decode.go
+++ b/tools/fxadmin/core/decode/decode.go
@@ -106,6 +106,18 @@ func configFromBlock(marshaledBlock []byte) (*cb.Config, error) {
 		return nil, errors.Wrap(err, "failed to unmarshal config payload")
 	}
 
+	if payload.GetHeader() == nil {
+		return nil, errors.New("block is not a config block: payload has no header")
+	}
+	channelHeader, err := protoutil.UnmarshalChannelHeader(payload.GetHeader().GetChannelHeader())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal channel header")
+	}
+	if channelHeader.GetType() != int32(cb.HeaderType_CONFIG) {
+		return nil, errors.Newf("block is not a config block: channel header type is %s, want %s",
+			cb.HeaderType(channelHeader.GetType()), cb.HeaderType_CONFIG)
+	}
+
 	configEnvelope, err := protoutil.UnmarshalConfigEnvelope(payload.Data)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to unmarshal config envelope")
--- a/tools/fxadmin/core/decode/decode.go
+++ b/tools/fxadmin/core/decode/decode.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/hyperledger/fabric-x-common/protolator"
 	"github.com/hyperledger/fabric-x-common/protoutil"
-	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/utils"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/helpers"
 )
 
 var logger = flogging.MustGetLogger("fxadmin.decode")
@@ -44,7 +44,7 @@ func New() *Handler {
 func (*Handler) Run(blockPath, outputPath string) error {
 	logger.Debugf("decode: block=%s output=%s", blockPath, outputPath)
 
-	if err := utils.RequireDistinctOutput(outputPath, blockPath); err != nil {
+	if err := helpers.RequireDistinctOutput(outputPath, blockPath); err != nil {
 		return err
 	}
 
--- a/tools/fxadmin/core/decode/decode.go
+++ b/tools/fxadmin/core/decode/decode.go
@@ -5,29 +5,29 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 // Package decode implements the `fxadmin decode` command, which converts a
-// binary config block into human-readable JSON. It reproduces the logic of
-// `configtxlator proto_decode --type common.Block`.
+// binary config block into the human-editable JSON of the common.Config it
+// carries. It reproduces `configtxlator proto_decode --type common.Block`
+// followed by extracting the embedded common.Config, so the output is the shape
+// `fxadmin compute-update` consumes: an admin edits this JSON and feeds the
+// original and edited copies back to compute-update.
 package decode
 
 import (
 	"bytes"
 	"io"
 	"os"
-	"reflect"
 
 	"github.com/cockroachdb/errors"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/reflect/protoreflect"
-	"google.golang.org/protobuf/reflect/protoregistry"
 
 	"github.com/hyperledger/fabric-x-common/protolator"
+	"github.com/hyperledger/fabric-x-common/protoutil"
 	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/utils"
 )
 
 var logger = flogging.MustGetLogger("fxadmin.decode")
-
-const blockMessageType protoreflect.FullName = "common.Block"
 
 // Handler executes the decode command.
 type Handler struct{}
@@ -37,10 +37,10 @@ func New() *Handler {
 	return &Handler{}
 }
 
-// Run implements `fxadmin decode`. It decodes the binary config block at
-// blockPath and writes its JSON rendering to outputPath. The output is written
-// only after the block decodes successfully, so a malformed block never
-// clobbers an existing destination.
+// Run implements `fxadmin decode`. It reads the binary config block at
+// blockPath, extracts the embedded common.Config, and writes its JSON rendering
+// to outputPath. The output is written only after the block decodes
+// successfully, so a malformed block never clobbers an existing destination.
 func (*Handler) Run(blockPath, outputPath string) error {
 	logger.Debugf("decode: block=%s output=%s", blockPath, outputPath)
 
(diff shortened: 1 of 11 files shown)


tools/fxadmin/core/update/compute_update.go
```diff
@@ -28,8 +34,51 @@ func New() *Handler {
...
+func encodeConfig(path string) (*cb.Config, error) {
+	input, err := os.Open(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open config %q", path)
+	}
+	defer func() { _ = input.Close() }()
+
+	config := &cb.Config{}
+	if err := protolator.DeepUnmarshalJSON(input, config); err != nil {
+		return nil, errors.Wrapf(err, "failed to decode config %q", path)
```
HagarMeir: why not use %s here?

tools/fxadmin/core/update/compute_update.go
```diff
@@ -28,8 +34,51 @@ func New() *Handler {
...
+	current, err := encodeConfig(currentPath)
+	if err != nil {
+		return err
+	}
+	modified, err := encodeConfig(modifiedPath)
+	if err != nil {
+		return err
+	}
+
+	configUpdate, err := update.Compute(current, modified)
```
HagarMeir, line 76: The resulting `ConfigUpdate` never has its `ChannelId` set. Every other caller of `update.Compute` in this repo sets it immediately afterward — `configtxgen/encoder.go:501` (`updt.ChannelId = channelID`), `configtxlator/rest/configtxlator_handlers.go:67`, and the real `configtxlator compute_update` CLI (`cmd/configtxlator/main.go:226`, whose `--channel_id` flag is `Required()`). A `ConfigUpdate` with an empty `ChannelId` cannot be applied: Fabric validates that the `ConfigUpdate.ChannelId` matches the enclosing envelope’s channel header, so an update produced by this command is unusable as-is. The command also takes no channel argument to supply one. Consider adding a channel-id argument/flag and setting `configUpdate.ChannelId` before marshaling.

tools/fxadmin/core/update/compute_update.go
```diff
@@ -6,20 +6,26 @@ SPDX-License-Identifier: Apache-2.0
 
 // Package update implements the `fxadmin compute-update` command, which
 // computes the ConfigUpdate delta between the original and modified
-// configuration JSON.
+// configuration JSON. It reproduces the logic of `configtxlator proto_encode
+// --type common.Config` on each input followed by `configtxlator
+// compute_update`: each JSON config is encoded to a common.Config, and the
+// delta between the two is written as a marshaled common.ConfigUpdate.
```
HagarMeir, line 12: This doc comment (and the one on `Run`) claims the command reproduces `configtxlator proto_encode` followed by `configtxlator compute_update`, but the reproduction is incomplete: `configtxlator compute_update` requires a `--channel_id` and always sets `ConfigUpdate.ChannelId`, while this implementation never sets it. Either update the doc to state the divergence or (better) close the gap by setting `ChannelId`.

tools/fxadmin/core/update/compute_update.go
```diff
@@ -28,8 +34,51 @@ func New() *Handler {
...
+	configUpdate, err := update.Compute(current, modified)
+	if err != nil {
+		return errors.Wrap(err, "failed to compute config update")
+	}
+
+	out, err := proto.Marshal(configUpdate)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal config update")
+	}
+	if err := os.WriteFile(outputPath, out, 0o600); err != nil {
```
HagarMeir, line 86: No guard against the output path colliding with an input. The sibling `decode` command added `requireDistinctPaths` (`tools/fxadmin/core/decode/decode.go`) precisely to stop the output silently clobbering its source. Here `compute-update` has two inputs (`currentPath`, `modifiedPath`) and does an unconditional `os.WriteFile`, so e.g. `fxadmin compute-update current.json modified.json --output modified.json` silently overwrites the user's `modified.json` with the binary ConfigUpdate. Worth generalizing the `decode` safety check and sharing it here rather than omitting it.

tools/fxadmin/core/update/compute_update_test.go
```diff
@@ -7,18 +7,123 @@ SPDX-License-Identifier: Apache-2.0
...
 )
 
-// TestRunNotImplemented asserts the compute-update command is a skeleton that
-// reports "not implemented". Replace with a delta-computation test once the
-// command is implemented.
-func TestRunNotImplemented(t *testing.T) {
+// TestRunComputesConfigUpdate asserts `fxadmin compute-update` encodes both
+// JSON configs to common.Config and writes the ConfigUpdate delta between them:
+// the changed value lands in the write set with its version bumped, and the
+// read set retains the original version.
```
HagarMeir: This comment misdescribes what the test asserts. Line 51 asserts `require.NotContains(t, configUpdate.GetReadSet().GetValues(), "HashingAlgorithm")` — i.e. the changed value is *absent* from the read set (correct: `update.Compute` puts changed leaf values only in the write set). The comment says the read set 'retains the original version', which is the opposite, and would mislead a future maintainer about ConfigUpdate read/write-set semantics.

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: A flag with the current block is added, and the channel ID will be taken from there.

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: Added a comment that the channel ID is not part of common.Config, so it is read
from the current config block supplied to Run.

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: done

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: channel id was added

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: done

tools/fxadmin/core/update/compute_update_test.go
MayRosenbaum: channel id was added

tools/fxadmin/core/update/compute_update.go
```diff
@@ -28,8 +38,78 @@ func New() *Handler {
...
+// common.Config, mirroring `configtxlator proto_encode --type common.Config`.
+func encodeConfig(path string) (*cb.Config, error) {
+	input, err := os.Open(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open config %q", path)
+	}
+	defer func() { _ = input.Close() }()
+
+	config := &cb.Config{}
+	if err := protolator.DeepUnmarshalJSON(input, config); err != nil {
```
HagarMeir, line 123: `compute-update` consumes `common.Config` JSON here, but no fxadmin command *produces* that shape. The natural pipeline — `fxadmin ledger config` → `fxadmin decode` → edit → `compute-update` — breaks, because `decode` emits `common.Block`-shaped JSON (top-level `header`/`data`/`metadata`), not a `common.Config`. `DeepUnmarshalJSON` into `*common.Config` then rejects the block's top-level fields and `Run` fails with a confusing "failed to decode config". Consider either accepting block-shaped JSON (extract the embedded `Config`) or adding a command that emits `common.Config` JSON, so the pipeline is completable within fxadmin.

tools/fxadmin/core/helpers/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/cockroachdb/errors"
+)
+
+// RequireDistinctOutput reports an error if outputPath resolves to the same
+// file as any of inputPaths, so writing the output never silently overwrites
+// one of the command's own inputs. Paths are compared after resolving to
+// absolute form, so distinct strings that name the same file are still caught.
+func RequireDistinctOutput(outputPath string, inputPaths ...string) error {
+	outputAbs, err := filepath.Abs(outputPath)
```
HagarMeir, line 21: This compares paths with `filepath.Abs` only, which does not resolve symlinks. If `outputPath` is a symlink (or lies under a symlinked directory) that resolves to one of the `inputPaths`, the abs strings differ, the guard passes, and the subsequent `os.WriteFile` follows the link and overwrites the input — defeating the "never overwrite its own source" invariant this function's docstring promises. Consider resolving with `filepath.EvalSymlinks` (falling back to `Abs` when the path doesn't yet exist) before comparing.

tools/fxadmin/core/update/compute_update.go
```diff
@@ -28,8 +38,78 @@ func New() *Handler {
...
+	channelID, err := channelIDFromBlock(currentBlockPath)
+	if err != nil {
+		return err
+	}
+
+	configUpdate, err := update.Compute(current, modified)
+	if err != nil {
+		return errors.Wrap(err, "failed to compute config update")
+	}
+	configUpdate.ChannelId = channelID
```
HagarMeir, line 80: Two related gaps around the channel ID stamped here:

1. **No validation** — `channelIDFromBlock` accepts any block and an empty channel ID without error. A block whose first envelope has an empty `ChannelId` (or a non-CONFIG block, despite the "config block" help text) yields `ChannelId=""` silently, deferring failure to a much more opaque downstream submission error. An up-front non-empty check would fail fast.

2. **No consistency check** — with three independent file args, a user can pass `current.json` from channel A but a `--current-block` from channel B. The delta is computed from A's versions but stamped with B's channel, producing a wrong-channel `ConfigUpdate` with no warning. Worth at least documenting the requirement, or validating the block's channel against the config source.

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: Good catch!
The decode command should take a `common.Block `and produce a JSON representation
of `common.Config`. This will be fixed in this PR.

tools/fxadmin/core/helpers/helpers.go
MayRosenbaum: very rare, I dont think we should support it

tools/fxadmin/core/update/compute_update.go
MayRosenbaum: Added ChannelId="" check.
Regarding the second point - Fabric-X is designed as a single channel, so an incorrect channel ID will be detected during submission. I added it in the doc above the function.

tools/fxadmin/core/helpers/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+	"path/filepath"
+
+	"github.com/cockroachdb/errors"
+)
+
+// RequireDistinctOutput reports an error if outputPath resolves to the same
+// file as any of inputPaths, so writing the output never silently overwrites
+// one of the command's own inputs. Paths are compared after resolving to
+// absolute form, so distinct strings that name the same file are still caught.
+func RequireDistinctOutput(outputPath string, inputPaths ...string) error {
```
HagarMeir, line 20: This guards only output-vs-input; it doesn't check the input paths against each other. `compute-update` calls `RequireDistinctOutput(output, current, modified, block)`, so accidental input aliasing gets no early, clear diagnostic: passing the same file for `current.json` and `modified.json` degrades to `update.Compute`'s opaque "no differences detected", and passing the binary block where a JSON config is expected degrades to a low-level `DeepUnmarshalJSON` decode error — rather than a message naming the aliased argument. Consider also rejecting duplicate/aliased inputs up front.

HagarMeir: maybe you can add this in the next PR?

tools/fxadmin/core/decode/decode.go
```diff
@@ -55,60 +55,64 @@ func (*Handler) Run(blockPath, outputPath string) error {
...
-	if err := protolator.DeepMarshalJSON(output, msg); err != nil {
-		return errors.Wrap(err, "failed to encode output")
+
+	configEnvelope, err := protoutil.UnmarshalConfigEnvelope(payload.Data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal config envelope")
 	}
 
-	return nil
+	if configEnvelope.GetConfig() == nil {
```
HagarMeir, line 126: This relies solely on `GetConfig() != nil` and never checks the channel-header type is `HeaderType_CONFIG`. protobuf unmarshalling is field-number-based and lenient, so a non-config envelope whose payload happens to decode into a `ConfigEnvelope` with a non-nil `Config` will slip through here and `decode` will emit misleading JSON instead of a clear "not a config block" error. Extracting the `ChannelHeader` and checking `Type == int32(cb.HeaderType_CONFIG)` up front would reject non-config blocks deterministically.

tools/fxadmin/core/decode/decode.go
```diff
@@ -55,60 +55,64 @@ func (*Handler) Run(blockPath, outputPath string) error {
...
+		return nil, errors.New("block does not carry a config: config envelope has no config")
```
HagarMeir, line 127: Behavior-removal worth an explicit confirm: before this PR `fxadmin decode` rendered any `common.Block` (header/data/metadata) to JSON; now a normal/endorser or genesis block hard-errors here and produces no output. This is intentional given the command's redefinition around config blocks, but if any workflow or habit relied on `decode` to inspect an arbitrary block (block number, metadata, transactions), that path is now gone. Please confirm no caller depended on the general decoder — otherwise consider keeping a fallback that renders the raw block when it isn't a config block.

MayRosenbaum: if there is no config in the block, then decode cannot output anything.
the purpose of decode is to output a JSON of the config

tools/fxadmin/core/helpers/helpers.go
MayRosenbaum: No need to add this, if the content of the input is similar, then `update.Compute` (line 76 at compute_update.go) will result in: "no differences detected between original and updated config"

tools/fxadmin/core/decode/decode.go
MayRosenbaum: I added a check

HagarMeir (approved):

PR #154, opened 2026-08-16
[fxadmin] tx command merge and endorse

MayRosenbaum:
#### Type of change

- New feature

#### Description

TX endorse and merge commands in fxadmin.

#### Related issues

### DIFF
--- a/tools/fxadmin/core/tx/tx.go
+++ b/tools/fxadmin/core/tx/tx.go
@@ -9,10 +9,20 @@ SPDX-License-Identifier: Apache-2.0
 package tx
 
 import (
+	"bytes"
 	"fmt"
+	"os"
 
 	"github.com/cockroachdb/errors"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/hyperledger/fabric-x-common/common/util"
+	"github.com/hyperledger/fabric-x-common/msp"
+	"github.com/hyperledger/fabric-x-common/protoutil"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/signer"
+	"github.com/hyperledger/fabric-x-common/tools/fxadmin/core/user"
 )
 
 var logger = flogging.MustGetLogger("fxadmin.tx")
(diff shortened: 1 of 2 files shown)


tools/fxadmin/core/tx/tx.go
```diff
@@ -29,16 +39,138 @@ func New() *Handler {
...
-	return fmt.Errorf("tx merge: %w", errNotImplemented)
+
+	merged := &cb.ConfigUpdateEnvelope{}
+	seen := make(map[string]struct{})
+	for _, path := range inputPaths {
+		env, err := readConfigUpdateEnvelope(path)
+		if err != nil {
+			return err
+		}
+		if merged.ConfigUpdate == nil { // first iteration
```
HagarMeir: 🟡 **Minor**: using `merged.ConfigUpdate == nil` as the "first iteration" sentinel conflates *not yet set* with *set to empty*.

If the first input's `ConfigUpdate` is empty (`nil`/zero bytes), `merged.ConfigUpdate` stays `nil`, so the **next** input silently becomes the reference and the mismatch check never runs against that first empty input.

**Why:** a sentinel value that can also be a legitimate value is fragile. In practice `ConfigUpdate` is never legitimately empty (compute-update rejects "no differences" and endorse signs real bytes), so this is robustness rather than an exploitable bug.

**Recommendation (optional):** track first-seen explicitly — e.g. seed `merged.ConfigUpdate` from the first envelope before the loop, or use a `first bool` flag / the loop index instead of inferring it from `nil`.

HagarMeir (commented):
## Review Summary — fxadmin `tx merge` / `tx endorse` (+ compute-update)

Clean, well-scoped, and well-tested. Nothing blocking — only Minor/Nit robustness observations below.

### ✅ Verified correct

- **Signing convention.** `signConfigUpdate` signs `util.ConcatenateBytes(SignatureHeader, ConfigUpdate)` — byte-for-byte identical to the canonical producer (`tools/configtxgen/encoder.go:643`) and verifier (`protoutil/signeddata.go:53`, `Data = SignatureHeader || ConfigUpdate`). So endorsements produced here verify against channel-config policies downstream.
- **Signing the raw input bytes** rather than a re-marshaled `ConfigUpdate` is the right call — it guarantees every endorser signs byte-identical content and sidesteps non-deterministic proto map ordering in multi-party endorsement. Worth preserving against future "cleanup".
- **Merge dedup by `SignatureHeader.Creator`** — one signature per signer regardless of nonce, which is what threshold/majority config policies expect. Not verifying signatures at merge time is consistent with upstream Fabric (verification is the orderer's job against channel policies).
- Error handling (`cockroachdb/errors`, no `pkg/errors`), single `flogging` logger, license headers, `0o600` perms, and reuse of `protoutil`/`common/util`/`protolator`/`configtxlator/update` are all compliant. Tests are `testify` + table-driven + `t.Parallel()`, no new Ginkgo, and compare proto by unmarshaling fields (no `require.Equal`-on-proto). DCO signed on all commits.

### Additional Nits (not inlined)

- 🔵 `Endorse` accepts a zero-byte input as a valid (empty) `ConfigUpdate`, since `proto.Unmarshal` succeeds on empty bytes. Very low value to guard (the CLI requires an `ExistingFile`, and a downstream apply would reject an empty update) — noting only.
- 🔵 `Merge` with an empty `inputPaths` slice writes an empty envelope. Only reachable by calling the exported handler directly; the CLI arg is `Required().Strings()` (≥1). A one-line guard would make the handler self-defending if reused elsewhere. Optional.

No proto/ASN.1/config-schema/doc changes required — scope is tight.

tools/fxadmin/core/tx/tx.go
```diff
@@ -29,16 +39,138 @@ func New() *Handler {
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to read config update %q", inputPath)
+	}
+	if err = proto.Unmarshal(configUpdate, &cb.ConfigUpdate{}); err != nil {
+		return errors.Wrapf(err, "failed to unmarshal config update %q", inputPath)
+	}
+
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return err
```
HagarMeir: use errors.Wrapf here?

tools/fxadmin/core/tx/tx.go
```diff
@@ -29,16 +39,138 @@ func New() *Handler {
...
+	}
+	admin, err := signer.New(config.MSP.LocalMspID, config.MSP.LocalMspDir)
+	if err != nil {
+		return err
```
HagarMeir: use errors.Wrapf here?

tools/fxadmin/core/tx/tx.go
```diff
@@ -29,16 +39,138 @@ func New() *Handler {
...
+	}
+	if err = proto.Unmarshal(configUpdate, &cb.ConfigUpdate{}); err != nil {
+		return errors.Wrapf(err, "failed to unmarshal config update %q", inputPath)
+	}
+
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return err
+	}
+	admin, err := signer.New(config.MSP.LocalMspID, config.MSP.LocalMspDir)
```
HagarMeir, line 62: need to check for nil values before?

tools/fxadmin/core/tx/tx.go
```diff
@@ -29,16 +39,138 @@ func New() *Handler {
...
+}
+
+// signConfigUpdate produces the admin's ConfigSignature over the config update.
+// The signature covers SignatureHeader||ConfigUpdate.
+func signConfigUpdate(admin msp.SigningIdentity, configUpdate []byte) (*cb.ConfigSignature, error) {
+	sigHeader, err := protoutil.NewSignatureHeader(admin)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create signature header")
+	}
+	signatureHeader := protoutil.MarshalOrPanic(sigHeader)
```
HagarMeir: why use MarshalOrPanic here and not just return an error instead of panic?

tools/fxadmin/core/tx/tx.go
MayRosenbaum: checked in `user.LoadConfig(configPath)` by the function `validate`

tools/fxadmin/core/tx/tx.go
MayRosenbaum: no real reason, fixed

tools/fxadmin/core/tx/tx.go
MayRosenbaum: This exposed a small bug, fixed

HagarMeir (commented):
## Review — last commit (`tx endorse and merge`)

Reviewed the latest commit in isolation. It incorporates the prior round's feedback and is clean — no blocking issues.

### ✅ Prior findings addressed
- **Merge first-iteration anchoring** — the `merged.ConfigUpdate == nil` sentinel is replaced with an explicit `if i == 0`, so an empty first `ConfigUpdate` still anchors the mismatch comparison. The new `"empty first config update still anchors the comparison"` test is a solid regression guard (fails under the old logic, passes under the new).
- **Merge non-verification** is now documented ("Signatures are not verified here; their validity is checked against channel policy at submission").

### ✅ Improvements in this commit
- `signConfigUpdate` uses checked `proto.Marshal(sigHeader)` instead of `MarshalOrPanic` — no panic path in library-style code.
- `Endorse` now wraps each origin error with context (`failed to load user configuration %q`, `failed to load admin signing identity for MSP %q`, `failed to endorse config update %q`) instead of bare `return err`.

### ✅ Core still correct
- Signature covers `ConcatenateBytes(SignatureHeader, ConfigUpdate)`, matching the producer/verifier convention (`tools/configtxgen/encoder.go:643`, `protoutil/signeddata.go:53`). Endorse signs the raw input bytes (deterministic across endorsers); Merge dedups by `SignatureHeader.Creator`.
- Tests are `testify` + table-driven + `t.Parallel()` + `t.Helper()`, compare proto by unmarshaling fields (no `require.Equal`-on-proto), and add no Ginkgo.

### 🔵 Nit (optional)
In `Endorse`, a couple of the outer wraps slightly duplicate context the callee already provides — e.g. `errors.Wrapf(err, "failed to load user configuration %q", configPath)` wraps `user.LoadConfig`, which already emits `failed to read user configuration file %q`. Harmless module-level context; not worth changing on its own.

Net: resolves the one Minor from the previous round and tightens error handling. Looks good to merge.

HagarMeir (approved):

PR #156, opened 2026-08-17
[fxadmin] tx command prepare

MayRosenbaum:
#### Type of change

- New feature

#### Description

Prepare command in fxadmin.

#### Related issues

### DIFF
--- a/tools/fxadmin/core/tx/tx.go
+++ b/tools/fxadmin/core/tx/tx.go
@@ -177,10 +177,50 @@ func appendSignatures(merged, env *cb.ConfigUpdateEnvelope, seen map[string]stru
 	return nil
 }
 
-// Prepare implements `fxadmin tx prepare`.
+// Prepare implements `fxadmin tx prepare`. It reads the endorsed
+// common.ConfigUpdateEnvelope at inputPath and wraps it in a common.Envelope
+// whose channel header is of type CONFIG_UPDATE, signed by the submitting
+// client identity described by the configuration YAML at configPath (a channel
+// writer, which may be one of the admins). The marshaled envelope is written to
+// outputPath, ready for submission to the routers.
 func (*Handler) Prepare(inputPath, configPath, outputPath string) error {
 	logger.Debugf("tx prepare: input=%s config=%s output=%s", inputPath, configPath, outputPath)
-	return fmt.Errorf("tx prepare: %w", errNotImplemented)
+
+	env, err := readConfigUpdateEnvelope(inputPath)
+	if err != nil {
+		return err
+	}
+	configUpdate := &cb.ConfigUpdate{}
+	if err = proto.Unmarshal(env.GetConfigUpdate(), configUpdate); err != nil {
+		return errors.Wrapf(err, "failed to unmarshal config update in endorsement %q", inputPath)
+	}
+
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to load user configuration %q", configPath)
+	}
+	client, err := signer.New(config.MSP.LocalMspID, config.MSP.LocalMspDir)
+	if err != nil {
+		return errors.Wrapf(err, "failed to load submitting client signing identity for MSP %q", config.MSP.LocalMspID)
+	}
+
+	tx, err := protoutil.CreateSignedEnvelope(
+		cb.HeaderType_CONFIG_UPDATE, configUpdate.GetChannelId(), client, env, 0, 0)
+	if err != nil {
+		return errors.Wrap(err, "failed to create configuration transaction")
+	}
+
+	out, err := proto.Marshal(tx)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal configuration transaction")
+	}
+	if err := os.WriteFile(outputPath, out, 0o600); err != nil {
+		return errors.Wrapf(err, "failed to write output %q", outputPath)
+	}
+
+	logger.Infof("prepared configuration transaction from %s signed by %s to %s",
+		inputPath, config.MSP.LocalMspID, outputPath)
+	return nil
 }
 
 // Submit implements `fxadmin tx submit`.
--- a/tools/fxadmin/core/tx/tx_test.go
+++ b/tools/fxadmin/core/tx/tx_test.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	notImplemented = "not implemented"
-	adminYAML      = "admin.yaml"
+	notImplemented   = "not implemented"
+	adminYAML        = "admin.yaml"
+	missingInputCase = "missing input file"
 )
 
 // TestHandlerNotImplemented asserts every not-yet-implemented tx subcommand is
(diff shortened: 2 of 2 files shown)


tools/fxadmin/core/tx/tx.go
```diff
@@ -29,22 +39,184 @@ func New() *Handler {
...
+// outputPath.
 func (*Handler) Endorse(inputPath, configPath, outputPath string) error {
 	logger.Debugf("tx endorse: input=%s config=%s output=%s", inputPath, configPath, outputPath)
-	return fmt.Errorf("tx endorse: %w", errNotImplemented)
+
+	configUpdate, err := os.ReadFile(inputPath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read config update %q", inputPath)
+	}
+	if err = proto.Unmarshal(configUpdate, &cb.ConfigUpdate{}); err != nil {
```
HagarMeir, line 54: 🔵 **Nit:** the decoded message is intentionally discarded here — this is a *parse sanity-check* (we sign the raw `configUpdate` bytes so every endorser signs identical content). That intent isn't obvious from the code, and protobuf parsing is lenient (many byte strings decode into unknown fields), so this only rejects gross garbage, not a semantically empty update. A short comment would help a future reader:

```suggestion
	// Sanity-check that the input parses as a ConfigUpdate; we sign the raw bytes
	// below so every endorser signs identical content.
	if err = proto.Unmarshal(configUpdate, &cb.ConfigUpdate{}); err != nil {
```

tools/fxadmin/core/tx/tx.go
```diff
@@ -29,22 +39,184 @@ func New() *Handler {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to sign config update")
+	}
+	return &cb.ConfigSignature{
+		SignatureHeader: signatureHeader,
+		Signature:       signature,
+	}, nil
+}
+
+// Merge implements `fxadmin tx merge`. It reads several marshaled
```
HagarMeir, line 109: 🟡 **Minor:** `Merge` unions and de-duplicates signatures but does **not** cryptographically verify them — which is the right call for an offline preparation tool (validity is enforced against channel policy on submission, and an invalid signature simply fails at the orderer; it can't authorize an unintended change). Worth one line in the doc comment making that trust boundary explicit, so a future reader doesn't mistake `merge` for a verification step. e.g. append: *"Signatures are not verified here; their validity is checked against channel policy at submission."*

HagarMeir (commented):
## Review Summary — fxadmin tx command prepare

Thanks for this — it lands the full offline config-update pipeline (`compute-update` → `tx endorse` → `tx merge` → `tx prepare`) cleanly, reusing existing plumbing (`protoutil`, `configtxlator/update`, `protolator`, `common/util`) instead of reinventing it. **No Major findings.**

### Verified correct
- **Signing convention matches the verifier.** `signConfigUpdate` signs `SignatureHeader‖ConfigUpdate`, which is exactly what `protoutil.ConfigUpdateEnvelopeAsSignedData` reconstructs (`protoutil/signeddata.go:53`). ✅
- **Endorse signs the raw input bytes** (not a re-marshaled message), so every endorser signs byte-identical content — required for multi-org merge. ✅
- **Prepare** preserves the inner `ConfigUpdate` (a `bytes` field) and all endorser signatures through re-marshal, and takes the channel ID from the decoded update. ✅
- **Merge dedup** keys on `SignatureHeader.Creator` — distinct admins kept, same identity deduped. ✅
- Compliance: license headers, `cockroachdb/errors` + `%w`, `flogging`, `0o600` perms, table-driven parallel tests, no Ginkgo, no proto `require.Equal`. ✅

### Notes (nothing blocking)
- 2 inline comments below (1 Minor doc suggestion, 1 Nit).
- **Nit (scope/description):** the title says "prepare" but the PR actually implements four commands (compute-update, endorse, merge, prepare). Since the commit body becomes the PR description verbatim, consider enumerating them.
- No `.proto`/ASN.1/`configtx.yaml`/`go.mod` changes → no compat, config-chain, or dependency obligations.

Comment-only per review policy.

tools/fxadmin/core/tx/tx.go
HagarMeir: https://github.com/hyperledger/fabric-x-common/pull/154

tools/fxadmin/core/tx/tx.go
HagarMeir: https://github.com/hyperledger/fabric-x-common/pull/154

tools/fxadmin/core/tx/tx.go
MayRosenbaum: This protects a case in which the config update is not a valid `configUpdate` input.

tools/fxadmin/core/tx/tx.go
MayRosenbaum: added in #154

tools/fxadmin/core/tx/tx.go
```diff
@@ -177,10 +177,50 @@ func appendSignatures(merged, env *cb.ConfigUpdateEnvelope, seen map[string]stru
...
+		return err
+	}
+	configUpdate := &cb.ConfigUpdate{}
+	if err = proto.Unmarshal(env.GetConfigUpdate(), configUpdate); err != nil {
+		return errors.Wrapf(err, "failed to unmarshal config update in endorsement %q", inputPath)
+	}
+
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return err
```
HagarMeir: wrap?

tools/fxadmin/core/tx/tx.go
```diff
@@ -177,10 +177,50 @@ func appendSignatures(merged, env *cb.ConfigUpdateEnvelope, seen map[string]stru
...
+	}
+	client, err := signer.New(config.MSP.LocalMspID, config.MSP.LocalMspDir)
+	if err != nil {
+		return err
```
HagarMeir: wrap?

HagarMeir (commented):
## Re-review — clean after re-scope ✅

Re-reviewed following the rebase. The PR is now cleanly scoped to the **`tx prepare`** command (endorse/merge/compute-update split into #154, now in `main`), and the description names the command + links the orderer issue — the earlier scope/description nit is addressed.

**`Prepare` implementation:** correct. Reads the endorsed `ConfigUpdateEnvelope`, decodes `ConfigUpdate` only to read the channel ID, and wraps via `protoutil.CreateSignedEnvelope(CONFIG_UPDATE, channelID, client, env, 0, 0)`. The inner `ConfigUpdate` (a `bytes` field) and every endorser signature survive the re-marshal — `TestPrepareMergedEndorsement` proves the 2-signature case. Error handling (`cockroachdb/errors` + `%w`), `flogging`, `0o600`, and license header are all in order.

**Tests:** `TestPrepare` / `TestPrepareMergedEndorsement` / `TestPrepareErrors` — table-driven, `t.Parallel`, `require`, no Ginkgo, proto fields compared individually (no `require.Equal` on whole messages), reusing the #154 helpers instead of duplicating. The `missingInputCase` const is a tidy in-scope touch.

My two earlier inline findings were on code that moved to #154 and were addressed there. **No new findings — LGTM.** (Comment-only per review policy.)

HagarMeir (approved):

PR #158, opened 2026-08-20
add block_hash and prev_block_hash to TxEventBatch

arner:
#### Type of change

- New feature

#### Description

Adds two fields to the TxEventBatch, the return object from StreamAllTransactions of the committer's Notifier. See the related issue on fabric-x-committer for more details.

#### Related issues

https://github.com/hyperledger/fabric-x-committer/issues/773

### DIFF
--- a/api/committerpb/notify.proto
+++ b/api/committerpb/notify.proto
@@ -81,6 +81,12 @@ message TxEventBatch {
 
     // List of transaction events (may be filtered based on request).
     repeated TxEvent events = 2;
+
+    // The hash of the block from which these transactions originated.
+    bytes block_hash = 3;
+
+    // The hash of the previous block.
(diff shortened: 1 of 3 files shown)


api/committerpb/notify.proto
```diff
@@ -81,6 +81,12 @@ message TxEventBatch {
 
     // List of transaction events (may be filtered based on request).
     repeated TxEvent events = 2;
+
+    // The hash of the block from which these transactions originated.
+    bytes block_hash = 3;
+
+    // The hash of the previous block.
+    bytes prev_block_hash = 4;
```
cendhu, line 89: With these additions, we should rename TxEventBatch to BlockEvent. We can make the change in this PR itself.

cendhu (approved):

PR #159, opened 2026-08-23
[fxadmin] tx command submit

MayRosenbaum:
#### Type of change
New feature

#### Description
Submit command in fxadmin.

#### Related issues
https://github.com/hyperledger/fabric-x-orderer/issues/1067

### DIFF
--- a/tools/fxadmin/core/client/client.go
+++ b/tools/fxadmin/core/client/client.go
@@ -77,11 +77,11 @@ func load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error
 func LoadFromFiles(configPath, blockPath string, csp bccsp.BCCSP) (*Client, error) {
 	config, err := user.LoadConfig(configPath)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to load user configuration %q", configPath)
 	}
 	block, err := readBlock(blockPath)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to load config block %q", blockPath)
 	}
 	return load(config, block, csp)
 }
--- a/tools/fxadmin/core/client/client.go
+++ b/tools/fxadmin/core/client/client.go
@@ -45,8 +45,8 @@ type Client struct {
 	tlsCertHash     []byte
 }
 
-// Load builds a Client from the user configuration and the config block.
-func Load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error) {
+// load builds a Client from the user configuration and the config block.
+func load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error) {
 	ordererConnInfo, err := ordererconn.Load(block, csp)
 	if err != nil {
 		return nil, err
@@ -69,6 +69,34 @@ func Load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error
 		signer:          signingIdentity,
 		tlsCertHash:     certHash,
 	}, nil
+}
+
+// LoadFromFiles builds a Client from the user configuration YAML at configPath
+// and the config block at blockPath, using csp to build the channel config
+// bundle.
+func LoadFromFiles(configPath, blockPath string, csp bccsp.BCCSP) (*Client, error) {
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return nil, err
+	}
+	block, err := readBlock(blockPath)
+	if err != nil {
+		return nil, err
+	}
+	return load(config, block, csp)
+}
+
+// readBlock reads and unmarshals a protobuf config block from path.
+func readBlock(path string) (*cb.Block, error) {
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read config block %q", path)
+	}
+	block, err := protoutil.UnmarshalBlock(content)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal config block %q", path)
+	}
+	return block, nil
 }
 
 // ChannelID returns the channel the client targets.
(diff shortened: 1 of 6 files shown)


tools/fxadmin/core/tx/tx.go
```diff
@@ -223,10 +227,69 @@ func (*Handler) Prepare(inputPath, configPath, outputPath string) error {
...
+		return err
+	}
+
+	statuses, err := cl.BroadcastToAllRouters(envelope)
+	if err != nil {
+		return errors.Wrap(err, "failed to broadcast configuration transaction")
+	}
+	reportBroadcast(statuses)
+	return nil
+}
```
HagarMeir, line 260: 🔴 **Major:** `Submit` returns `nil` here regardless of how many routers acknowledged the transaction. Since `BroadcastToAllRouters` now only errors when *no routers are configured*, an all-reject or all-unreachable broadcast still exits 0 — the failures are only `Warnf`-logged (and suppressed above WARN level). For a CLI the exit code is the primary contract, so a script/CI step would treat "nothing submitted" as success.

**Why:** dropping the old "error if *any* router rejects" is right for a multi-router network where one ack suffices, but success on *zero* acks is the opposite extreme.

**Recommendation:** enforce a minimum-ack threshold (natural one: at least one):
```suggestion
	reportBroadcast(statuses)
	if countAcks(statuses) == 0 {
		return errors.Newf("configuration transaction not acknowledged by any of %d routers", len(statuses))
	}
	return nil
```
If a different contract is intended (all must ack, or deliberately fire-and-forget), please state it in the PR description and a code comment so the exit-code behavior is intentional.

HagarMeir (commented):
## Review — `[fxadmin] tx command submit`

Thanks for wiring up `tx submit` — the `RouterStatus` design is clean, the refactor into `client.LoadFromFiles` colocates the file loading sensibly, and the tests are strong (table-driven, in-process gRPC ack/reject stubs, `t.Parallel`/`t.Helper`/`t.Cleanup`, no sleeps, no new Ginkgo). Everything is idiomatic w.r.t. `cockroachdb/errors`, `flogging`, and `protoutil` reuse, and `csp: factory.GetDefault()` matches `ledger.New()`. Verified repo-wide that nothing external calls `Load` or `BroadcastToAllRouters`, so unexporting `load` and reshaping the broadcast signature are safe.

One **Major** correctness concern and one **Nit** inline. COMMENT-only.

### 🔴 Major — `tx submit` exits 0 even when zero routers accepted the transaction
`BroadcastToAllRouters` now returns a non-nil error only when *no routers are configured*; all rejections and unreachable routers are carried in `statuses`, `reportBroadcast` merely `Warnf`/`Infof`-logs them, and `Submit` then returns nil unconditionally. So if every router rejects (`BAD_REQUEST`) or is unreachable, the command still exits 0 — a scripted/CI caller sees success when nothing was submitted, and the only failure signal is a `Warnf` line that a higher log level suppresses. Moving off the old "error if *any* router rejects" is reasonable for a multi-router network, but success on *zero* acks is the opposite extreme. Details inline.

### Testing
No case asserts the exit-code behavior on an all-reject / all-unreachable broadcast — exactly the gap above. If the threshold change is made, please add a `Submit` case pointed at a rejecting/unreachable router that asserts a non-nil error.

### Docs
No config/proto/make/structure change — no doc updates required.

tools/fxadmin/core/tx/tx.go
```diff
@@ -223,10 +227,69 @@ func (*Handler) Prepare(inputPath, configPath, outputPath string) error {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal configuration transaction %q", path)
+	}
+	return envelope, nil
+}
+
+// countAcks returns the number of routers that acknowledged the transaction,
+// those whose status carries no error.
+func countAcks(statuses []client.RouterStatus) int {
+	acked := 0
```
HagarMeir: 🔵 **Nit:** `countAcks` is currently a single-call helper (only `reportBroadcast` uses it), which we'd normally inline. This becomes moot if the Major above is addressed — `Submit` would then be a second caller — so only worth acting on if it stays single-use.

tools/fxadmin/core/tx/tx.go
MayRosenbaum: `reportBroadcast` prints all statuses from all routers with a log of how many acknowledgments have been received. no need for additional log.

tools/fxadmin/core/tx/tx.go
MayRosenbaum: let's keep it as it is

tools/fxadmin/core/client/client.go
```diff
@@ -71,6 +71,34 @@ func Load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error
 	}, nil
 }
 
+// LoadFromFiles builds a Client from the user configuration YAML at configPath
+// and the config block at blockPath, using csp to build the channel config
+// bundle.
+func LoadFromFiles(configPath, blockPath string, csp bccsp.BCCSP) (*Client, error) {
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return nil, err
```
HagarMeir: wrap

tools/fxadmin/core/tx/tx.go
```diff
@@ -223,10 +227,69 @@ func (*Handler) Prepare(inputPath, configPath, outputPath string) error {
...
+	envelope, err := protoutil.UnmarshalEnvelope(content)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal configuration transaction %q", path)
+	}
+	return envelope, nil
+}
+
+// countAcks returns the number of routers that acknowledged the transaction,
+// those whose status carries no error.
+func countAcks(statuses []client.RouterStatus) int {
```
HagarMeir: countAcks should come after reportBroadcast that uses it

tools/fxadmin/core/tx/tx_test.go
```diff
@@ -40,16 +44,67 @@ func TestHandlerNotImplemented(t *testing.T) {
...
 		t.Parallel()
 		require.ErrorContains(t, h.Send("endorsed.pb", adminYAML, "current.pb"), notImplemented)
 	})
 }
 
+// TestSubmitErrors asserts that `tx submit` reports readable errors, before any
+// network access, for a missing transaction file, a transaction that is not a
+// well-formed envelope, a missing configuration file, and an unreadable config
+// block.
+func TestSubmitErrors(t *testing.T) {
```
HagarMeir, line 62: kind of funny that there is no submit test without errors 
but I understand most of the checks are done by the client tests

tools/fxadmin/core/tx/tx.go
MayRosenbaum: moved

tools/fxadmin/core/tx/tx_test.go
MayRosenbaum: I added a test, but yes, this is covered very good in the client_test.go, I think the test of submit is redundant

HagarMeir (commented):
## Follow-up review (after "address comments" ×2)

Thanks for the quick turnaround — the revision addresses the main gap well.

**Addressed** ✅
- **Testing:** `TestSubmit` (real ARMA config-block fixture via `cryptogen`, three routers: ack / reject / unreachable) and `TestSubmitNoRouters` now exercise the full `Submit` path end-to-end — exactly the coverage that was missing. High-fidelity fixture.
- **Shared test helpers:** extracting `StartBroadcastServer`/`FreeAddress` into `tools/fxadmin/core/client/test` is a clean de-dup, and it follows the existing repo pattern (`tools/test/tree.go` is likewise `package test` importing `testing` outside a `_test.go` file). License header + package doc present.

**Two small residual points** (inline) — one is my original Major, now a documented design decision I'd just like an explicit call on; the other is a nit the revision introduced. Nothing blocking. COMMENT-only.

tools/fxadmin/core/tx/tx.go
HagarMeir, line 260: Following up on my earlier Major: the new doc comments + `TestSubmit` make the intent clear — `Submit` deliberately does **not** fail on *individual* router rejections, which is the right call for a multi-router network. 👍

The one edge that remains is **zero acks**: if *every* router rejects or is unreachable, `Submit` still returns nil (exit 0) and the failure is only `Warnf`-logged. `TestSubmit` always keeps one acking router, so this case isn't covered. For a CLI whose exit code drives scripts/CI, "nothing was accepted" reading as success is the part worth an explicit decision.

No need to change the individual-rejection behavior — just consider whether `countAcks(statuses) == 0` should be an error (and a matching test). If you'd rather keep it fire-and-forget, a one-line note in the doc comment saying zero-ack is intentionally success would close this out.

tools/fxadmin/core/client/client.go
```diff
@@ -71,6 +71,34 @@ func Load(config *user.Config, block *cb.Block, csp bccsp.BCCSP) (*Client, error
...
 }
 
+// LoadFromFiles builds a Client from the user configuration YAML at configPath
+// and the config block at blockPath, using csp to build the channel config
+// bundle.
+func LoadFromFiles(configPath, blockPath string, csp bccsp.BCCSP) (*Client, error) {
+	config, err := user.LoadConfig(configPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to load user configuration %q", configPath)
+	}
```
HagarMeir, line 81: 🔵 **Nit:** this wrap (and the `failed to load config block %q` one just below) duplicates context that the callees already provide. `user.LoadConfig` returns `failed to read user configuration file %q: …` and `readBlock` returns `failed to unmarshal config block %q: …`, so the chain ends up repeating the same path and a near-identical phrase, e.g. `failed to load config block "x": failed to unmarshal config block "x": …`. Since the inner errors already carry the path + operation, returning them unwrapped (as before) reads cleaner. Minor — feel free to keep if you prefer the outer framing.

MayRosenbaum: I got a comment to wrap, so lets keep it

tools/fxadmin/core/tx/tx.go
MayRosenbaum: added a note

tock-ibm (approved):

PR #164, opened 2026-08-26
[api] Add checkpoint feedback transport

### LINKED ISSUE #163: [api] Add checkpoint feedback transport
Checkpoint verification needs a control signal separate from per-transaction commit status. `api/committerpb/status.proto` currently carries only transaction statuses, although `TxStatusBatch` is already the response envelope on both the VC→coordinator and coordinator→sidecar paths.

Add an optional checkpoint feedback message that:

- expresses `PROCEED`, `HOLD`, and `HALT` without changing transaction `Status`;
- carries checkpoint transaction ID, referenced snapshot block number, and HALT divergence reason;
- documents that PROCEED releases HOLD, HOLD/HALT block only new intake while fetched work drains, and HALT is terminal;
- remains additive and reusable on both existing stream paths.

cendhu:
#### Type of change

- New feature

#### Description

- Add optional checkpoint feedback to `TxStatusBatch` for the VC→coordinator and coordinator→sidecar paths.
- Define PROCEED, HOLD, and HALT signals with checkpoint transaction ID, block number, and diagnostic reason.
- Regenerate Go protobuf bindings.

- Keep transaction commit `Status` unchanged.
- Verify with `make proto`, `make lint-proto`, `go test ./api/committerpb -race`, and `go build ./...`.
- Repository has no `make build` target.

#### Related issues

- resolves #163

### DIFF
--- a/api/committerpb/status.proto
+++ b/api/committerpb/status.proto
@@ -12,9 +12,37 @@ package committerpb;
 
 import "api/committerpb/ref.proto";
 
-// A batch of TXs' status.
+// A batch of TXs' status and optional checkpoint feedback.
 message TxStatusBatch {
     repeated TxStatus status = 1;
+    // Checkpoint feedback from the VC to the coordinator. The coordinator forwards the same message to the sidecar.
+    // If absent, ordinary per-TX status is authoritative.
+    optional CheckpointFeedback checkpoint_feedback = 2;
+}
+
+// CheckpointFeedback controls checkpoint intake while local snapshot hash verification completes.
+message CheckpointFeedback {
+    // Signal controls new intake. HOLD and HALT allow already-fetched work to drain.
+    enum Signal {
+        // No checkpoint feedback. Ordinary per-TX status is authoritative.
+        SIGNAL_UNSPECIFIED = 0;
+        // Local snapshot hash is computed and verified. Releases a prior HOLD and resumes intake.
+        PROCEED = 1;
+        // Local snapshot hash is not yet computed. Blocks new intake while already-fetched work drains.
+        HOLD = 2;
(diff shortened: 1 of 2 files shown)


api/committerpb/status.proto
```diff
@@ -12,9 +12,37 @@ package committerpb;
...
 import "api/committerpb/ref.proto";
 
-// A batch of TXs' status.
+// A batch of TXs' status and optional checkpoint feedback.
 message TxStatusBatch {
     repeated TxStatus status = 1;
+    // Checkpoint feedback from the VC to the coordinator. The coordinator forwards the same message to the sidecar.
+    // If absent, ordinary per-TX status is authoritative.
+    optional CheckpointFeedback checkpoint_feedback = 2;
+}
```
liran-funaro, line 21: `TxStatusBatch` is only used in the committer internally. It should be moved to the committer.
This is unrelated to this PR, but since you modify it, I suggest addressing this issue in the committer first, then removing this redundant proto from the common repo.

api/committerpb/status.proto
```diff
@@ -12,9 +12,37 @@ package committerpb;
...
+
+// CheckpointFeedback controls checkpoint intake while local snapshot hash verification completes.
+message CheckpointFeedback {
```
liran-funaro, line 24: Is this used outside of the committer? Reported to users/admins?
If not, this should not be in the common as well.

cendhu:
Superseded by https://github.com/hyperledger/fabric-x-common/pull/169. TxStatusBatch and checkpoint feedback are committer-internal transport types, so the envelope moves to fabric-x-committer as servicepb.TxStatusBatch instead of gaining a checkpoint_feedback field here.

PR #166, opened 2026-08-26
[blkstorage] Do not build tx index information no index will read

### LINKED ISSUE #165: [blkstorage] Do not build tx index information no index will read
The effect is measured on the Fabric-X Committer, whose sidecar is the block store user in question:
this is part of the umbrella hyperledger/fabric-x-committer#798, and pairs with
hyperledger/fabric-x-committer#784, which is what configures a store to index by block number alone.

`serializeBlock` extracts a transaction ID for every envelope and builds a `txindexInfo` and a
`locPointer` for each, whatever the store is configured to index. Only two indexes read any of it: the
txID index needs the ID and the offset, and the blockNum-tranNum index needs the offset. A store
configured with neither — which is what a committer sidecar with the transaction ID index disabled
does, since it indexes by block number alone — pays for all of it and reads none.

The transaction ID is the expensive half: `GetOrComputeTxIDFromEnvelope` unmarshals the whole envelope
and its payload header per transaction. On a nineteen-machine cluster driving 10,000-transaction blocks,
a CPU profile of the sidecar put `addDataBytesAndConstructTxIndexInfo` at 7.4% of the whole process and
about 88% of `appendBlock`. Because the append is one serialized goroutine, that was the pipeline's
ceiling: the stage ran at a 100% duty cycle — 22.2 ms of a 22.2 ms per-block budget — on a machine
sitting at 18% CPU with its disk 4% busy.

Have `serializeBlock` ask the index what it will read. Append 22.2 → 7.1 ms per block, which lifted the
sidecar's own ceiling from ~451,000 tps. The serialized bytes are unchanged for every combination.

liran-funaro:
#### Type of change

- Test update

#### Description

- `serializeBlock` takes an `indexNeeds`, and `blockIndex.serializationNeeds` derives it from
  the configured attributes, so the decision lives next to `indexBlock`, which is the code that
  consumes the result. Neither the txIDs nor the offsets are produced unless an index reads them
  back.
- Split the offsets out into `constructTxIndexInfo`, which runs only when an index wants them.
  It derives each envelope's span from the envelope lengths rather than observing them mid-append
  (`protowire.AppendBytes` writes a length prefix plus the payload, so `protowire.SizeBytes`
  gives the same value the old `len(buf) - offset` did), and takes one backing array each for the
  infos and their locations instead of two allocations per transaction.
- Pre-allocate the serialization buffer to `serializedBlockSize(block)`, the exact length the
  writers produce, instead of growing it by append.
- Rename `addHeaderBytes` / `addMetadataBytes` to `appendHeaderBytes` / `appendMetadataBytes` and
  put the buffer first, matching the `protowire.Append*` shape. This diverges from upstream
  Fabric's names.
- Drop the txID from the blockNum-tranNum `Debugf` in `indexBlock`: it is only extracted when the
  txID index is configured, so it would print empty for a store that indexes by block number.
- Add `TestSerializeBlockIndexNeeds` (the serialized bytes are identical for every combination,
  and what was not asked for is absent), `TestBlockIndexSerializationNeeds` (a table over the
  attribute combinations), `TestSerializedBlockSizeIsExact`, and `BenchmarkSerializeBlock` /
  `BenchmarkSerializedBlockSize`.

The txID is the expensive half: `GetOrComputeTxIDFromEnvelope` unmarshals the whole envelope and
its payload header per transaction. On a 19-machine committer cluster driving 10,000-transaction
blocks, a CPU profile of the sidecar put block serialization (then
`addDataBytesAndConstructTxIndexInfo`, now inlined into `serializeBlock`) at 7.4% of the whole
process and about 88% of `appendBlock`, of which more than half was that unmarshal. Because the
append is one serialized goroutine, that was the pipeline's ceiling: the stage ran at a 100% duty
cycle on a machine sitting at 18% CPU with its disk 4% busy.

A store configured with neither index -- which is what a Fabric-X committer sidecar with
`disable-tx-id-index` set is, since it indexes by block number alone -- paid for all of it and
read none of it.

Measured on a 10,000-transaction block of 300-byte transactions, serialization with no tx index
goes from 8.7 ms and 21.1 MB allocated to 1.0 ms and 3.9 MB; with all indexes, from 25.9 ms to
14.9 ms. Roughly half of that is skipping the work and half is the pre-allocation: buffer growth
reallocates about 34 times per block and copies the whole buffer each time, whereas the size pass
is 14 us and zero allocations on the same block.

Verified: `make lint` reports 0 issues and `go test ./common/ledger/blkstorage/...` passes. The
strongest check on the computed offsets is the pre-existing `require.Equal(info, infoFromBB)` in
`testSerializedBlockInfo` -- `extractSerializedBlockInfo` re-derives the locations by parsing the
serialized bytes, so any drift in the spans fails it.

#### Related issues

- resolves #165

### DIFF
--- a/common/ledger/blkstorage/block_serialization.go
+++ b/common/ledger/blkstorage/block_serialization.go
@@ -26,14 +26,40 @@ type txindexInfo struct {
 	loc  *locPointer
 }
 
-func serializeBlock(block *common.Block) ([]byte, *serializedBlockInfo) {
-	var buf []byte
-	info := &serializedBlockInfo{}
-	info.blockHeader = block.Header
-	info.metadata = block.Metadata
-	buf = addHeaderBytes(block.Header, buf)
-	info.txOffsets, buf = addDataBytesAndConstructTxIndexInfo(block.Data, buf)
-	buf = addMetadataBytes(block.Metadata, buf)
+// indexNeeds says which of the per transaction facts an index will actually read back out
+// of a serialized block. Serialization sits on the critical path of every append, and both
+// of these cost real work at block scale: the offsets are an allocation per transaction,
+// and a txID is a full envelope unmarshal each. Neither is produced unless an index reads
+// it -- see blockIndex.indexBlock, where txOffsets is touched only under
+// IndexableAttrTxID or IndexableAttrBlockNumTranNum, and txID only under the former.
+type indexNeeds struct {
+	txOffsets bool
+	txIDs     bool
+}
+
+// allIndexNeeds asks for everything serialization can produce. For callers with no index to
+// consult, which is what the block store's own tests and the read-back paths want.
+var allIndexNeeds = indexNeeds{txOffsets: true, txIDs: true}
+
+func serializeBlock(block *common.Block, needs indexNeeds) ([]byte, *serializedBlockInfo) {
+	buf := make([]byte, 0, serializedBlockSize(block))
+	buf = appendHeaderBytes(buf, block.Header)
+	buf = protowire.AppendVarint(buf, uint64(len(block.Data.Data)))
+	dataStart := len(buf)
+	for _, txEnvelopeBytes := range block.Data.Data {
+		buf = protowire.AppendBytes(buf, txEnvelopeBytes)
+	}
+	buf = appendMetadataBytes(buf, block.Metadata)
+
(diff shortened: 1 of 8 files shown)


cendhu (approved):
LGTM

PR #179, opened 2026-09-02
[BREAKING] [committerpb] Consolidate sidecar RPCs into SidecarService

cendhu:
#### Type of change

- Breaking change

#### Description

- Replace `BlockQueryService` and `Notifier` with one `SidecarService` while preserving their method names and request/response messages.
- Add `DeleteDBCloneForSnapshot(DeleteDBCloneForSnapshotRequest)` to the unified service so the committer can implement snapshot-clone administration without another service migration.
- Keep Fabric block delivery on the Fabric `protos.Deliver` service for Fabric client wire compatibility.

The Delete RPC identifies the snapshot with `tx_id`, which is the `_snapshot` namespace row key. The proto documents why the resource-oriented AIP-135 Delete rules are disabled rather than renaming this field to `name`.


liran-funaro (approved):
I suggest also adding the regular block stream (Fabric style block stream).
But make the signature more like the `StreamBlocks` API. I.e., a request instead of a client to server stream, and a server to client stream.
And we can also remove the envelope wrapper (use directly the seek request) as we will add the ACL later.

cendhu:
> I suggest also adding the regular block stream (Fabric style block stream). But make the signature more like the `StreamBlocks` API. I.e., a request instead of a client to server stream, and a server to client stream. And we can also remove the envelope wrapper (use directly the seek request) as we will add the ACL later.

I agree. We can do this in a separate PR as the consolidation was mainly done to introduce DeleteClone API.

PR #181, opened 2026-09-07
[applicationpb] Encode the transaction digest without reflection

### LINKED ISSUE #180: [applicationpb] The signing digest is built by reflection
`TxNamespace.ASN1Marshal` builds the digest that every transaction is signed over and that every
verifier recomputes. It translates the namespace into an intermediate struct tree and hands that to
`encoding/asn1.Marshal`, which is reflection driven.

A CPU profile of a load generator driving 451,826 tps against a nineteen-machine committer put
`encoding/asn1.Marshal` at **14.31% of process CPU**, of which 80.5% — 11.5% of the whole process — is
this digest. `asn1.makeField` is 85.9% of the marshal. For comparison, the elliptic curve
multiplication in the same signature is 21.79%, so **building the message to sign costs over half of
signing it**, and `translate`'s own intermediate allocations are a further ~1.4%.

This is on both sides of the wire. The committer's verifier
(`utils/signature/verify.go` in fabric-x-committer) recomputes the same digest per transaction, so the
cost is paid once when signing and again on every validation.

### What to do

Encode the DER directly instead of reflecting over a struct tree. Measured at the transaction shape a
committer evaluation generates — two read-write operations, 32-byte keys and values, one metadata item:

| read-writes | `encoding/asn1` | direct | speedup |
|---|---|---|---|
| 2 | 7,510 ns/op, 2,384 B, 61 allocs | **1,512 ns/op, 1,200 B, 9 allocs** | **5.0x** |
| 8 | 18,687 ns/op | 4,594 | 4.1x |
| 64 | 124,103 ns/op | 31,796 | 3.9x |
| 512 | 940,608 ns/op | 231,277 | 4.1x |

Verified end to end on a 39-machine deployment with a real Arma ordering service: 480,057 tps at 460 ms
mean and 700 ms p99 with **zero aborts**, which is the correctness check that matters — the verifier
recomputes the digest independently of the signer, so a single byte of disagreement fails every
transaction.

### Constraints on any implementation

**The output must stay byte-identical.** A change here invalidates every signature made by an older
build, so old and new must agree exactly. Go's `encoding/asn1` has two marshal rules that are easy to
miss:

- `optional` without a default omits a field only when it is `DeepEqual` to the type's zero value. For
  `Metadata [][]byte` that is `nil`, so a **non-nil empty slice still encodes as an empty SEQUENCE**. A
  `len(metadata) > 0` guard gets this wrong and produces a different digest. Reachability is currently
  low — protobuf decodes an empty repeated field to nil, and known producers build nil or a populated
  slice — but it is a silent digest divergence, not a benign one.
- `optional,default:-1` on the version fields omits only on an exact match, and `TxID`/`NamespaceID`
  carry explicit `utf8` tags.

`api/applicationpb/asn1_test.go` already has `FuzzASN1MarshalTxNamespace` and a `generateTxNs` helper,
so the safety net is a differential assertion inside `requireASN1Marshal` comparing the new encoder to
`encoding/asn1` byte for byte, keeping the reflection path until the fuzzer has run against it. Note
the existing harness always passes a two-element metadata slice, so nil and empty need a separate
explicit test.

**Watch large values.** A nested-buffer approach that builds each sub-sequence into its own
`bytes.Buffer` and copies it into its parent pays for the payload three times. On a fixture whose keys
run to 1 MB that measured **4.7x slower** than `encoding/asn1` and allocated 5.5 MB against 1.25 MB,
even while being 5x faster at small values. Sizing the output in one pass, or pre-sizing the child
buffers, avoids the regression.

Related: hyperledger/fabric-x-committer#798 (the evaluation this came out of) and
hyperledger/fabric-x-committer#812 (the same reflection cost in the signature's own DER, already
fixed there with `cryptobyte`).

liran-funaro:
#### Type of change

- Test update

#### Description

- Add `TxNamespace.QuickASN1Marshal`, which writes the DER of `asn1_tx_schema.asn` directly. It
  sizes the whole encoding first (`tlvSize`, `readSize`, `readWriteSize`, `writeSize`,
  `positiveIntegerSize`), allocates that exact length once, then appends each field, so the result
  is one allocation and every payload byte is copied exactly once. `ASN1Marshal` is unchanged and
  remains the reference the new encoder is checked against.
- Mark `metadata` `OPTIONAL` in `asn1_tx_schema.asn`. The schema already did not match what
  `ASN1Marshal` produces: the field carries `asn1:"optional"`, so `encoding/asn1` omits it for a
  nil slice. The schema now says so.
- Check the two encoders against each other everywhere the marshalling is already tested:
  `requireASN1Marshal` (in `asn1_test_exports.go`, so the loadgen's use of it covers this too)
  requires byte equality and unmarshals both results, which puts every case in `txTestCases` and
  every `FuzzASN1MarshalTxNamespace` input on both paths.
- Add `TestMetadataEdgeCases`, for the one shape the fuzz target cannot reach: it always passes a
  two-element metadata slice, so nil, empty-non-nil, and a single empty element are never compared.
  A nil metadata is absent and a non-nil empty one is an empty SEQUENCE, which is what
  `encoding/asn1` does for an `optional` field with no default.
- Add `BenchmarkASN1Marshal` over the shapes the doc comment quotes.

Medians of `BenchmarkASN1Marshal` over `-count=5`:

| shape | `encoding/asn1` | `QuickASN1Marshal` |
|---|---|---|
| 2 read-writes, 32-byte keys and values | 7,195 ns, 61 allocs | **237 ns, 1 alloc** |
| 512 read-writes, 32-byte keys and values | 887,108 ns, 8,477 allocs | **26,460 ns, 1 alloc** |
| 2 read-writes, 1 MB keys and values | 2,433,735 ns, 61 allocs | **2,212,311 ns, 1 alloc** |

The first row is the shape a committer evaluation generates, and is 30x. The last row is a wash:
both encoders are dominated by copying the four megabytes, and that case's spread over nine runs is
20-26%, wider than the difference between them, so all it really shows is the allocation count.

The three sizing helpers lean on `math/bits`: `bits.Len64(value)/8 + 1` is a positive INTEGER's
length including the leading zero byte that keeps a value with its top bit set positive, since a bit
length that is a multiple of eight is exactly the case that fills the top byte. Both that and the
`(bits.Len(uint(l)) + 7) / 8` used for a length's byte count were checked against the loops they
replace exhaustively over the first 2^22 and 2^24 values plus the 64-bit edges.

Verified: `make lint` reports 0 issues on this change, `go test ./api/applicationpb/...` passes,
and `-fuzz FuzzASN1MarshalTxNamespace` ran 1.55M executions across two sessions with no difference
between the two encoders.

#### Related issues

- resolves #180


api/applicationpb/asn1_quick.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return size
+}
+
+// writeSize returns the content size of a Write's sequence.
+func writeSize(w *Write) int {
+	return tlvSize(len(w.Key)) + tlvSize(len(w.Value))
+}
+
+// tlvSize returns the encoded size of a TLV whose value occupies valueSize bytes.
```
cendhu: expand the first occurrence of TLV -- tag length value triplet.

api/applicationpb/asn1_quick.go
```diff
@@ -0,0 +1,205 @@
...
+	var tmp [8]byte
+	binary.BigEndian.PutUint64(tmp[:], uint64(l))
+	size := byteLen(l)
+	// [1] long - We add the encoding size and mark the MSB to indicate we use the long format.
+	//nolint:gosec // size is at most 8, so the tag byte cannot overflow.
+	return append(append(buf, 0x80|byte(size)), tmp[len(tmp)-size:]...)
+}
+
+// lengthSize returns the number of bytes appendLength emits for the given length.
+func lengthSize(l int) int {
```
cendhu: called only once. Hence, inline.

api/applicationpb/asn1_quick.go
```diff
@@ -0,0 +1,205 @@
...
+func appendTLVHeader(buf []byte, tag byte, valueSize int) []byte {
+	return appendLength(append(buf, tag), valueSize)
+}
+
+// appendLength uses a variable size number according to the ASN.1 scheme.
+// The MSB indicates whether the short or long format is used.
+//   - [0] short  Single byte where the remaining bits encode the length (max 127).
+//   - [1] long   The remaining bits of the first byte encode the size of the encoded length,
+//     followed by the said number of bytes encoding the length in big-endian.
+func appendLength(buf []byte, l int) []byte {
```
cendhu: called only once. Inline.

cendhu (approved):
LGTM

api/applicationpb/asn1_quick.go
liran-funaro: Good find. ponytail didn't find this

api/applicationpb/asn1_quick.go
liran-funaro: Are you suggesting to improve the comment?

api/applicationpb/asn1_quick.go
liran-funaro: Good find. ponytail didn't find this

Issue #190, opened 2026-09-15
[committerpb] A snapshot that can never be hashed has no terminal state


cendhu:
Part of the database snapshot & checkpoint feature (see `specs/001-db-snapshot/rfc-db-snapshot.md` in fabric-x-committer), and a follow-up to #143.

## Summary

A snapshot that can never be hashed has no terminal state, so a single such snapshot stops an organization from ever taking another one.

## The gap

The validator-committer admits a new snapshot only once the latest `_snapshot` record is `CHECKPOINTED` (`REJECTED_SNAPSHOT_IN_PROGRESS` / `REJECTED_SNAPSHOT_NO_CHECKPOINT`, added in #143). A record parked at `FAILED` is retried forever, and there is no way out but editing the state database by hand.

There is no way to say "abandon this snapshot" — and it cannot be local state either: a node re-ingesting from genesis rebuilds the state at the snapshot height, would hash successfully, and would publish a digest for a snapshot the original organization abandoned.

## Proposal

An **abort snapshot transaction**, submitted by an administrator, ordered and endorsed on the ledger. In `fabric-x-common` this needs:

1. `SnapshotState.Status.ABORTED = 6` — terminal, meaning the snapshot will not be hashed and no checkpoint attests to it. Written only by the validator-committer when an abort transaction commits (the rule `CHECKPOINTED` already follows), never by the snapshot hasher: a hasher aborting on its own findings would let one organization abandon a snapshot another hashes successfully, and the two histories would diverge on re-ingestion.
2. Two statuses for an abort that names no abortable snapshot:
   - `MALFORMED_SNAPSHOT_INVALID_ABORT_KEY` — the read-write key does not decode as an abort key.
   - `REJECTED_SNAPSHOT_ABORT_NO_SUCH_SNAPSHOT` — a well-formed key naming a block that is not the snapshot awaiting its checkpoint, or one whose lifecycle is already closed (`CHECKPOINTED`/`ABORTED`).

   Both are per-transaction rejections rather than failures, because the submitter chose the block number and local state is not in question.
3. The abort write-key encoding (`SnapshotAbortKey` / `IsSnapshotAbortKey` / `BlockNumFromSnapshotAbortKey`). It belongs here, not in fabric-x-committer, because the only legitimate submitter is an administrator whose tooling depends on this module: a key only the committer could construct would make the transaction unbuildable by the one party allowed to send it.

[...]

PR #191, opened 2026-09-15
[committerpb] Add the abort snapshot transaction status and key encoding

cendhu:
#### Type of change

- New feature

#### Description

- `SnapshotState.Status` gains `ABORTED = 6`: terminal, meaning the snapshot will not be
  hashed. Written only by the validator-committer when an abort snapshot transaction
  commits, the rule `CHECKPOINTED` already follows, and never by the snapshot hasher.
- `MALFORMED_SNAPSHOT_INVALID_ABORT_KEY = 122`: an abort transaction's read-write key does
  not decode as an abort key.
- `REJECTED_SNAPSHOT_ABORT_NO_SUCH_SNAPSHOT = 123`: a well-formed abort key naming a block
  that is not the snapshot awaiting its checkpoint, or one whose lifecycle is already
  closed. Both are per-transaction rejections rather than failures, because the submitter
  chose the block number and local state is not in question.
- `MALFORMED_SNAPSHOT_NOT_MARKER_ONLY`'s comment is updated: a `_snapshot` TX is now valid
  in two shapes, a marker-only snapshot request and an abort carrying exactly one
  read-write.
- `SnapshotAbortKey(blockNum)` builds the `_snapshot` write key of an abort snapshot
  transaction: the `abort/` prefix followed by the block number, encoded order-preserving
  as a checkpoint key is.
- `IsSnapshotAbortKey` inspects the prefix only, so a prefixed key whose block number does
  not decode is rejected as a malformed abort rather than mistaken for a snapshot request.
- `BlockNumFromSnapshotAbortKey` decodes the block number and rejects trailing bytes rather
  than ignoring them: accepting a prefix would attribute the abort to the wrong snapshot.
- `snapshot.pb.go` and `status.pb.go` regenerated via `make proto`.

The key encoding lives in this module rather than in fabric-x-committer because the only
legitimate submitter of an abort is an administrator, whose tooling depends on this module.

#### Related issues

- resolves partly #190


api/committerpb/snapshot_abort_key.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/hyperledger/fabric-x-common/common/ledger/util"
+)
+
+// snapshotAbortKeyPrefix distinguishes an abort key from a snapshot record's key. Both
+// live in the `_snapshot` namespace table, and a record's key is a transaction ID, so
+// an abort must be recognizable from the key alone: that is what lets the commit path
+// tell an abort transaction from a snapshot request without reading the record it
+// names. A transaction ID is a hex-encoded digest, so it cannot contain '/'.
+var snapshotAbortKeyPrefix = []byte("abort/")
```
liran-funaro: The issue is that we never actually verify this.
If we did, we could decode the TX-ID into binary and reduce the write size to the DB and over the network.
This was flagged as an issue during the evaluation, but I dismissed it because we never enforce "hex-encoded digest".
If we want to rely on it, we must enforce it in the sidecar.

PR #194, opened 2026-09-18
added possibility to set a custom admin user in `crypto-config.yaml`

pasquale95:
#### Type of change

- New feature

#### Description

- Add `Admin` (bool, default `false`) to `UserSpec` in `crypto-config.yaml`, letting a declared user be generated with admin capabilities instead of a plain member — same folder path and naming convention as any other user (`users/<name>@<domain>`).
- With `EnableNodeOUs: false` (the common case), admin authority comes from admincerts membership rather than the cert's OU, so `overwriteAdminCert`/`overwriteNodesAdminCert` were generalized to `overwriteAdminCerts`/`syncAdminCerts`, which sync the *set* of admin certs (the default `Admin@<domain>` plus any user flagged `Admin: true`) into the org's and every node's `admincerts/` folder, instead of handling only the single pregenerated admin.
- `extendOrg` now also syncs the org-level admincerts (previously only node-level), so `cryptogen extend` picks up newly declared admins on existing networks.
- With `EnableNodeOUs: true`, the flag simply assigns the `admin` OU to the cert instead, consistent with how admin classification already works in that mode.
- Not setting `Admin` on a user preserves today's behavior exactly.

Verified with `go test ./...` and `golangci-lint run --new-from-rev=main`. Added `TestGenerateAdminUser` in `tools/cryptogen/org_crypto_test.go`, parametrized over `EnableNodeOUs` true/false, asserting the flagged user satisfies the MSP `ADMIN` role principal and a plain member does not.


tools/cryptogen/org_crypto.go
```diff
@@ -268,10 +250,14 @@ func (c *orgCryptoTree) generateUsers() []NodeSpec {
...
 	for _, spec := range s.Users.Specs {
+		ou := ClientOU
+		if spec.Admin {
+			ou = AdminOU
+		}
 		users = append(users, NodeSpec{
 			CommonName:         fmt.Sprintf("%s@%s", spec.Name, orgName),
 			PublicKeyAlgorithm: publicKeyAlg,
-			OrganizationalUnit: ClientOU,
+			OrganizationalUnit: ou,
```
mbrandenburger: Should we use `getPublicKeyAlg(spec.PublicKeyAlgorithm)` here to actually use the value as provided by the config?

tools/cryptogen/org_crypto.go
```diff
@@ -284,20 +270,49 @@ func (c *orgCryptoTree) generateUsers() []NodeSpec {
...
 	return nil
 }
 
-func (c *orgCryptoTree) overwriteAdminCert(adminCertsDir, adminUserName string) error {
-	adminCertPath := filepath.Join(adminCertsDir, adminUserName+"-cert.pem")
-	if _, err := os.Stat(adminCertPath); !os.IsNotExist(err) {
+func (c *orgCryptoTree) overwriteAdminCerts(adminCertsDir string, adminUserNames ...string) error {
+	if allAdminCertsExist(adminCertsDir, adminUserNames) {
 		return nil
 	}
```
mbrandenburger: I have the feeling that running `allAdminCertsExist` here will cause skipping rebuilding `admincerts` even if we are "removing a user". Should we just re-build here? WDYT?

mbrandenburger (commented):
Thank you @pasquale95 for working on this. Please see my comments below.

tools/cryptogen/org_crypto.go
pasquale95: Resolved.

tools/cryptogen/org_crypto.go
pasquale95: Resolved.

pasquale95:
Hi @mbrandenburger, both comments were valid. I've updated the code as per the PR comments.

mbrandenburger (approved):
LGTM

